PR #1, opened 2023-10-26
Smaller typos

martinknechtel:


martinknechtel:
Simon applied them directly.

Fannon:
Thanks for reporting! They're already fixed in the internal version and the fixes will be replicated here soon.

General note: PRs can be made here, but the proposals need to be backported to the internal version first.

PR #18, opened 2025-10-20
fix: Incorrect REUSE syntax

dellagustin-sap:
This PR fixes the syntax of the file `REUSE.toml`.
As it is, the repository is showing a non-compliant status on the REUSE badge:
screenshot of the specification repository showing the non-compliant status on the REUSE badge

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -7,5 +7,5 @@ SPDX-PackageComment = "The code in this project may include calls to APIs (\"API
 [[annotations]]
 path = "**"
 precedence = "aggregate"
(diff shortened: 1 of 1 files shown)


Fannon:
@dellagustin-sap now I already merged it, but I wonder if we should change this now that ORD belongs to Linux foundation. WDYT, @christianneu ?

dellagustin-sap:
Hi @Fannon , I'll wait for a reply from @christianneu, as I don't know if that changes with the donation to the foundation. I would assume that it does not hurt to keep using REUSE, but I it is likely that the copyright information must be updated.

PR #26, opened 2025-11-16
Fix Docusaurus image warnings in build

Fannon:


Fannon:
I don't think this works well, we'll have to live with docusaurus warnings - maybe docusaurus will fix this in the future by upgrading the affected library

PR #32, opened 2025-11-28
Add visibility to resource / capability definitions

Fannon:
### Added

- Added optional `visibility` to API, Event and Capability definitions
  - By default the definitions have the same visibility as the resource they belong to
  - This can be useful if e.g. a public resource has some metadata definitions that are internal only.
  - The visibility of a resource definition MUST be lower than the visibility of the resource it describes.


swennemers (commented):
generally fine, please add example

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ## [unreleased]
 
+### Added
+
+- Added optional `visibility` to API, Event and Capability definitions
```
maiargu:
```suggestion
- Added optional `visibility` to API Resource Definition, Event Resource Definition and Capability Definition
```

Not completely sure but what about `Exposed Entity Type` does that not need an optional `visibility` property?

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
...
+  - By default the definitions have the same visibility as the resource they belong to
+  - This can be useful if e.g. a public resource has some metadata definitions that are internal only.
+  - The visibility of a resource definition MUST be lower than the visibility of the resource it describes.
```
maiargu: `The visibility of a resource definition MUST be lower than the visibility of the resource it describes.`
Not sure I understand this fully, a valid example would probably help.

spec/v1/Document.schema.yaml
```diff
@@ -2504,6 +2504,15 @@ definitions:
           - /api/openApi.yaml
           - https://example.com/api/API_RFQ_PROCESS_SRV/$metadata
 
+      visibility:
+        <<: *visibility
+        description: |-
```
maiargu, line 2512: Is it possible to give `description` an anchor name and not duplicated it 3 times? It could run out of sync in future.

Fannon: yes, I try it

CHANGELOG.md
Fannon: I rephrased it. Is it better now?

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ## [unreleased]
 
+### Added
+
+- Added optional `visibility` to API Resource Definition, Event Resource Definition and Capability Definition
+  - By default the definitions have the same visibility as the resource they belong to
+  - The visibility of a resource definition MUST be lower (more restrictive) than the visibility of the resource it describes.
+    E.g. a public resource can have metadata definitions that are internal only.
```
maiargu: `E.g. a public resource can have metadata definitions that are internal only.`
can you pls add such an example to one of the example files?

CHANGELOG.md
maiargu: yes

CHANGELOG.md
swennemers:
```suggestion
    E.g. a public resource can declare to have some resource definitions that are internal while others are public. An internal resource can't set a resource definition to be public.
```

swennemers: I provided a suggestion below

spec/v1/Document.schema.yaml
```diff
@@ -2504,6 +2503,18 @@ definitions:
...
           - https://example.com/api/API_RFQ_PROCESS_SRV/$metadata
 
+      visibility: &definitionVisibility
+        <<: *visibility
+        description: |-
+          The visibility states who is allowed to "see" and access the resource definition, in case it differs from the resource visibility.
+
+          If not given, the resource definition has the same visibility as the resource it describes.
+          The visibility of a resource definition MUST be lower (more restrictive) than the visibility of the resource it describes.
+          E.g. a public resource can have metadata definitions that are internal only.
```
```suggestion
          E.g. a public resource can have metadata definitions that are internal only. An internal resource can't declare to have a public metadata definition.
```

swennemers (approved):
To be merged with 1.13

swennemers (approved):

PR #36, opened 2025-12-02
Add labels and correlationIDs to group and group type

Fannon:
Adding `labels` and `correlationIds` to group and group type. 
This allows to apply extensions / references to non ORD concepts via IDs, making the group concept more extensible.


spec/v1/Document.schema.yaml
```diff
@@ -3896,9 +3896,11 @@ definitions:
       description:
         <<: *description
 
-      # TODO: Add documentationLabels and more in the future? (Currently causing a bug)
-      # documentationLabels:
-      #   <<: *documentationLabels
+      labels:
+        <<: *labels
```
swennemers, line 4178: @Fannon , what is the use case for labels?

Fannon: If you want to add any extensions to a group or group type, you couldn't really do it without labels

PR #37, opened 2025-12-03
Fix PR Preview: Ensure PAT is used instead of GITHUB_TOKEN

pavelkornev:
Disable checkout credentials (`persist-credentials: false`) so the deployment action can push using the configured PAT instead of falling back to `github-actions[bot]`.


spec/v1/Document.schema.yaml
```diff
@@ -3,7 +3,7 @@ $id: https://open-resource-discovery.github.io/specification/spec-v1/interfaces/
 title: Ord Document
 description: |-
   The [ORD Document](../index.md#ord-document) object serves as a wrapper for the **ORD resources** and **ORD taxonomy** and adds further top-level information
-  that are specific to the document/the service it describes.
+  that are specific to the document/the service it describes.test
```
pavelkornev:
```suggestion
  that are specific to the document/the service it describes.
```

PR #40, opened 2025-12-08
Add members to Group and add Group Type restriction hints

Fannon:
### Added

- Added `members` assignment from a Group to ORD Resources
  - So far, ORD resources could only define their `partOfGroups` assignment, assuming they know which group (instances) they belong to
  - Sometimes not just the group taxonomy is defined externally, but also the group assignment. In this case, it is the group which knows which ORD resources are part of it (`members`), not vice versa
  - For an ORD aggregator this comes with a new challenge, as the relationship between group and its constituents can be defined from both sides
- Added `restrictDirection` to optionally restrict the direction of group assignments to either resource -> group or vice versa.
- Added `restrictResourceType` to optionally restrict group assignments to one or more ORD resource types.

@swennemers I'm least sure about `restrictDirection` if we really need this. Also we could leave out `restrictResourceType` for now. It might help with validation of correct use of taxonomy later, though.

### DIFF
--- a/docs/spec-v1/concepts/grouping-and-bundling.md
+++ b/docs/spec-v1/concepts/grouping-and-bundling.md
@@ -155,7 +155,7 @@ Example: Group Type: "Process", Group: ["Hire to Retire", "Travel to Reimburse"]
 
 Third, we need to assign the ORD resources to the groups, via:
 
-- `partOfGroup` assignment from ORD Resource to 0..n Groups
+- `partOfGroups` assignment from ORD Resource to 0..n Groups
 - `members` assignment from Group to 0..n ORD Resources
 
 E.g. a particular OData API for Employee Management can be part of both the "Employee Service" group (of Group Type "CSN Service") and the "Hire to Retire" group of Group Type "Process".
--- a/docs/spec-v1/concepts/grouping-and-bundling.md
+++ b/docs/spec-v1/concepts/grouping-and-bundling.md
@@ -163,7 +163,7 @@ E.g. a particular OData API for Employee Management can be part of both the "Emp
 Group Types and Groups can be locally defined by the ORD Provider ([system namespace](../../spec-v1/index.md#system-namespace)) or be defined globally ([authority namespace](../../spec-v1/index.md#authority-namespace)) if they are shared across different applications.
 
 Groups and even group types can be defined globally. If the Group Type is shared across different applications, it should have an [authority namespace](../index.md#authority-namespace).
-In this case it works like a global taxonomy with a predefined list of values.It's also possible for the application itself to define and assign its own Group Types (and instances).
+In this case it works like a global taxonomy with a predefined list of values. It's also possible for the application itself to define and assign its own Group Types (and instances).
 
 > The Group concept is correct choice when ORD resources need to be grouped through an explicit taxonomy, beyond the predefined taxonomy concepts from ORD (like Package). The advantage of groups over tags and labels is that their types and instance values can be defined, governed and listed.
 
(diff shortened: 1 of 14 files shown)


docs/spec-v1/concepts/grouping-and-bundling.md
```diff
@@ -145,16 +149,23 @@ They are optimized towards machine-readability and can be used to query, select
 The concept has three parts: The **Group Type** defines the semantics / meaning of a particular type of group assignments.
 An example would be to have a Group Type for a "part of an CSN Service" or "part of a Process".
```
swennemers:
```suggestion
An example would be to have a Group Type for a "CSN Service" or "Process".
```
that something is "part of", is the relationship property, not the type itself, isn't it.

docs/spec-v1/concepts/grouping-and-bundling.md
```diff
@@ -145,16 +149,23 @@ They are optimized towards machine-readability and can be used to query, select
...
 
-Second, the **Group** itself defines the actual group things can be assigned to.
-In the examples before, this would be the "Employee Service" or the "Hire to Retire" Process.
+Second, the **Group** itself defines the actual group instances that resources can be assigned to.
+
+Example: Group Type: "Process", Group: ["Hire to Retire", "Travel to Reimburse"]
+
+Third, we need to assign the ORD resources to the groups, via:
+
+- `partOfGroup` assignment from ORD Resource to 0..n Groups
```
swennemers:
```suggestion
- `partOfGroups` assignment from ORD Resource to 0..n Groups
```

docs/spec-v1/concepts/grouping-and-bundling.md
```diff
@@ -145,16 +149,23 @@ They are optimized towards machine-readability and can be used to query, select
...
+Group Types and Groups can be locally defined by the ORD Provider ([system namespace](../../spec-v1/index.md#system-namespace)) or be defined globally ([authority namespace](../../spec-v1/index.md#authority-namespace)) if they are shared across different applications.
 
-The Group Type could even be defined globally. If the Group Type is shared across different applications, it should have an [authority namespace](../index.md#authority-namespace).
-The Group Instances can potentially be globally defined, too. In this case it works like a global taxonomy with a predefined list of values.It's also possible for the application itself to define and assign its own Group Types and Instances.
+Groups and even group types can be defined globally. If the Group Type is shared across different applications, it should have an [authority namespace](../index.md#authority-namespace).
+In this case it works like a global taxonomy with a predefined list of values.It's also possible for the application itself to define and assign its own Group Types (and instances).
```
swennemers:
```suggestion
In this case it works like a global taxonomy with a predefined list of values. It's also possible for the application itself to define and assign its own Group Types (and instances).
```

spec/v1/Document.schema.yaml
```diff
@@ -3900,13 +3900,46 @@ definitions:
...
+            - "consumptionBundle"
+            - "apiResource"
+            - "eventResource"
+            - "capability"
+            - "entityType"
+            - "integrationDependency"
+            - "dataProduct"
+        x-introduced-in-version: "1.13.0"
+        x-feature-status: beta
+
```
swennemers, line 3934: I have the feeling, that we are getting too complicated with this and I would leave it out.

swennemers:
There is the probably overlapping requirement to introduce an explicit taxonomy/group assignment resource, as these assignments can have a different lifecycle than the ORD resource and also the taxonomy. So, I am wondering, if we should introduce members or if this would be better covered with a taxonomy/group assignment ORD resource.

spec/v1/Document.schema.yaml
Fannon: Yes, I have a similar feeling. I added it here because we discussed it. But we can decide to not do it for now, maybe keep it commented-out.

docs/spec-v1/concepts/grouping-and-bundling.md
Fannon: yes

PR #44, opened 2025-12-09
update dependencies

maiargu:
use spec-toolkit v0.5.0 with extensible enum documentation support


spec/v1/DocumentAPI.oas3.yaml
```diff
@@ -16,7 +16,7 @@ info:
     url: https://www.sap.com/
 externalDocs:
   description: "Open Resource Discovery Specification"
-  url: "https://open-resource-discovery.github.io/specification"
+  url: "https://openresourcediscovery.org"
```
maiargu, line 19: fyi @pavelkornev in case it changes to dashes usage, here is the place to adjust

Fannon (approved):

static/spec-v1/interfaces/ord-document-api/index.html
```diff
@@ -1,62 +0,0 @@
-<!doctype html>
```
maiargu, line 1: @pavelkornev can I delete this?

maiargu: added it back, it's used in an iframe

PR #46, opened 2025-12-11
Change domain name to open-resource-discovery.org

pavelkornev:


pavelkornev:
Wait until DNS records are through.

PR #48, opened 2025-12-15
Feat: Fetch banner url from banner-server

jurijat:
- A banner will be shown, when an external banner-server configuration is provided (`BANNER_SERVER_BASE_URL`) in `production` mode. Awaited content is an json object with an `url` attribute as string. 
- Mobile/Tablet mode banner fix


.github/workflows/main.yml
```diff
@@ -24,6 +24,8 @@ jobs:
 
       - name: Build
         run: npm run build
+        env:
+          BANNER_SERVER_BASE_URL: https://banner-server-mtls.cert.cfapps.sap.hana.ondemand.com
```
pavelkornev:
```suggestion
          BANNER_SERVER_BASE_URL: ${{vars.BANNER_SERVER_BASE_URL}}
```

Variable is already created.

static/js/custom.js
```diff
@@ -45,6 +46,127 @@ function __onNavScroll__() {
...
+  var apiUrl = getBannerApiUrl();
+  if (!apiUrl) {
+    console.debug(
+      "BANNER_SERVER_BASE_URL not configured, skipping banner fetch",
+    );
+    return;
+  }
+
+  try {
+    var data = await fetchWithRetry(apiUrl, 2, 1000);
```
pavelkornev, line 122: don't we cache results in the local storage to avoid re-fetches to banner server on every page load from the same user?

pavelkornev (approved):

PR #49, opened 2025-12-16
Add system-type perspective that is system-version independent

Fannon:
### Added

- Added `system-type` perspective to describe static metadata that is not version dependent.


pavelkornev:
@Fannon how will this perspective relate to `system-version`? E.g. provider can expose version-independent metadata and version-dependent metadata... shall the later put on top of the first? How the merge logic should work in this case?

`system type` < `system version` < `system instance`?

Do we need a separate perspective type at all here? Can't we reuse `system version` with some convention to the version, e.g. `0.0.0`? Or we can say that the system version is no longer mandatory field... if it's not specified, then it's not bound to any specific version.

static/spec-v1/interfaces/ums/MetadataType/externalordid.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: 'metadata-service.resource.api.sap/v6alpha1'
```
Fannon: Not sure why this is now checked in, shouldn't be part of this PR

docs/spec-v1/concepts/perspectives.md
```diff
@@ -22,12 +22,30 @@ Static metadata is known at design-time or deploy-time, when a new version of an
...
+In the case that the system is not versioned (continuous delivery) or resources do not relate to a system directly, the `system-type` perspective can be used.
+The `system-type` information apply no matter which system version or instance is used.
+
+The static perspective describes the shared metadata of all system instances (tenants) of the same system version or version independent.
 Either the metadata is always the same or it explicitly ignores the tenant-specific extensibility, configuration and any feature toggles - describing what's generic and shared.
 The advantage of static metadata is that is always available, there is no need to first provision tenants to get it. It is also a good integration contract for everything that is meant to work potentially with _any_ tenant.
 
 At SAP, we have the [SAP Business Accelerator Hub](https://api.sap.com/) that documents the static perspective.
 
+#### System Version Guidelines
```
Fannon, line 34: @swennemers  please review - is this also how you would propose it?

I'm not sure if we need an implied "latest" anymore (if system-version perspective, but no version is given). I'm afraid that starting with "latest" will become complicated quickly, so let's avoid it if not needed.

Fannon:
> @Fannon how will this perspective relate to `system-version`? E.g. provider can expose version-independent metadata and version-dependent metadata... shall the later put on top of the first? How the merge logic should work in this case?
> 
> `system type` < `system version` < `system instance`?
> 
> Do we need a separate perspective type at all here? Can't we reuse `system version` with some convention to the version, e.g. `0.0.0`? Or we can say that the system version is no longer mandatory field... if it's not specified, then it's not bound to any specific version.

Right now we don't merge at all, we just fall back a layer if one doesn't exist. So if system-instance layer exist, just use that. If it doesn't exist, fall back to entire system-version layer. If that doesn't exist, fall back to system-type layer. 

If we start with merging, this becomes much more complicated and some ORD Providers couldn't even do it because they don't keep track of this hierarchy on their own (what's an extension and what isn't).

NickyMateev (approved):

docs/spec-v1/concepts/perspectives.md
```diff
@@ -22,12 +22,30 @@ Static metadata is known at design-time or deploy-time, when a new version of an
...
+
+- If the system has explicit versions (e.g. "v1.0.1", "v2.0.0"), use these versions.
+- If the system does not use SemVer, it has to be converted so that SemVer conventions apply (conservatively).
+  - E.g. if the system uses simple incremental versions like "1", "2" or "2404, convert these to "1.0.0", "2.0.0", "2404.0.0" respectively.
+  - E.g. if the system uses date-based versions like "2024-01-15", convert these to "2024.1.15".
+- If the system is continuously delivered and does not have explicit versions at all, use the `system-type` perspective instead.
+  - However, consider that you may still have different releases rolled out over time and on different stages. If this causes a problem, use `system-version` with below conventions.
+  - Alternatively, use a fixed `1.0.0` version and overwrite it with every new release.
+- If the system has no versions, but we need to keep track of releases deployed on different environments or stages, use a synthetic versioning scheme:
+  - Use `1.0.0-DEV-SNAPSHOT`, `1.0.0-PROD-SNAPSHOT` or similar semver suffixes to indicate different stages or environments, where static metadata is identical.
```
swennemers:
```suggestion
```

swennemers: @Fannon  , based on latest discussion, do you want to keep this? I thought we don't want to have this mixed in the version?

docs/spec-v1/concepts/perspectives.md
```diff
@@ -13,7 +13,7 @@ An application or service can be described both from a static or a dynamic persp
 
 ### Static Perspective
```
swennemers, line 14: I am missing the `system-type` related thought in ## ORD Aggregator Considerations, 
e.g. system-type is always added to instance specific, and also to system-version...

docs/spec-v1/concepts/perspectives.md
Fannon: Hm, it's still a valid workaround - but it could also create even more confusions.

This would work if you're not in a facilitator situation. If we now introduce a `tier` explicitly, maybe that one should be used instead? But that would introduce more complications like the aggregator HAVING to understand it

Fannon:
> Static aggregators only describe the system-version perspective.
> 
> If both perspectives are described, it MUST only pick the system-version perspective

This needs to be updated

docs/spec-v1/concepts/perspectives.md
swennemers: +1

docs/spec-v1/concepts/perspectives.md
swennemers: +1 for providing it in the latest commit

swennemers (approved):

PR #52, opened 2025-12-16
Verify generated files are up to date

Fannon:
### Added 
- GitHub Action workflow to verify that `npm run generate` has been run, generated files are up-to-date
- Add `lefthook` to automatically run generate before committing (avoid the problem automatically)
- Created a dedicated `npm run clean` script, that does a more extensive cleanup and is more configurable. Removes `rimraf` dependency.


src/helper/clean.mjs
```diff
@@ -0,0 +1,133 @@
...
+const config = {
+  // Directories to remove entirely
+  removeDirs: [
+    'dist',
+    'build',
+    '.docusaurus',
+    'src/generated',
+    'logs',
+    'tmp'
+  ],
```
Fannon, line 11:
```suggestion
  removeDirs: [
    'dist',
    'build',
    'src/generated'
  ],
```

Probably this is enough

src/helper/clean.mjs
```diff
@@ -0,0 +1,133 @@
...
+  // Files to remove (absolute or workspace-relative)
+  removeFiles: [
+    // Log files in root
+    '*.log',
+    'npm-debug.log*',
+    // Static generated schema artifacts
+    'static/spec-v1/interfaces/Configuration.schema.json',
+    'static/spec-v1/interfaces/Document.schema.json'
+  ],
```
Fannon:
```suggestion
  removeFiles: [
    // Static generated schema artifacts
    'static/spec-v1/interfaces/Configuration.schema.json',
    'static/spec-v1/interfaces/Document.schema.json'
  ],
```

And here also just this

.github/workflows/main.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Installing Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: "22"
+          node-version: "24"
```
maiargu: probably you should not change this if package.json engine is 20

.github/workflows/verify-generated.yml
```diff
@@ -0,0 +1,57 @@
+# Verify Generated Files Workflow
+#
+# Purpose: Ensures all generated artifacts are committed.
+#
+# How: Installs deps, runs `npm run generate`, then checks for
+#      uncommitted changes. If any changes appear, the job fails
+#      and prints status + diff to guide fixes.
```
maiargu: delete comment

CONTRIBUTING.md
```diff
@@ -31,6 +31,28 @@ The following rule governs code contributions:
 - Contributions must be licensed under the [Apache 2.0 License](./LICENSE)
 - Due to legal reasons, contributors will be asked to accept a Developer Certificate of Origin (DCO) when they create the first pull request to this project. This happens in an automated fashion during the submission process. SAP uses [the standard DCO text of the Linux Foundation](https://developercertificate.org/).
 
+## Generated Artifacts and Review Process
+
+This repository checks in certain generated files (TypeScript types and UMS types). We intentionally commit these artifacts so pull requests clearly show the impact on those interfaces and downstream consumers during review.
+
+- What is generated: TypeScript interfaces in `dist/generated/spec/v1` and UMS types referenced under `spec/v1` and `src/generated/`.
```
maiargu:
```suggestion
- What is generated: TypeScript interfaces referenced unde `dist/generated/spec/v1` and `src/generated` and UMS types referenced under `spec/v1` and `src/generated/`.
```

.github/workflows/main.yml
Fannon: Not sure, if we bump engine, every project that has this as a dependency also needs to bump it. And there is actually just some types and data exported, which works with basically any node.js version I think. 

But we can still use newer versions in our pipelines

.github/workflows/verify-generated.yml
Fannon:
```suggestion
```

.github/workflows/main.yml
Fannon: Updated everything to be v22 consistently, which isn't the most recent LTS but still in maintenance

package.json
```diff
@@ -8,7 +8,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "engines": {
-    "node": ">=20.0.0",
+    "node": ">=22.0.0",
```
maiargu, line 11: re-execute npm install, this is not reflected in package-lock

maiargu: write the update as mention in the CHANGELOG file

Fannon: Done

.github/workflows/verify-generated.yml
Fannon: done

maiargu (approved):

PR #53, opened 2025-12-16
Ensure that minSystemVersion is a semver, as systemVersion already is

Fannon:
### Fixed

- **Breaking**: The `minSystemVersion` was not properly validated against semver, although the version it refers to (`describedSystemVersion.version`) is a mandatory semver.
  - Fixed the regex to properly validate against [Semantic Versioning 2.0.0](https://semver.org/) standard.
  - Added to documentation that this is an association to `SystemVersion.version`.
  - Introducing this as bugfix, as this property was presumed to be semver. This change just adds validation and explicit mentioning to ensure it.
  - If `minSystemVersion` is not a SemVer, it would be unclear how to do the version comparison anyway.


static/spec-v1/interfaces/ums/MetadataType/apiresource.yaml
```diff
@@ -94,9 +94,11 @@ spec:
     - name: 'disabled'
       type: 'boolean'
       description: "Indicates that this resource is currently not available for consumption at runtime, but could be configured to be so.\nThis can happen either because it has not been setup for use or disabled by an admin / user.\n\nIf the resource is not available in principle for a particular system instance, e.g. due to lack of entitlement, it MUST not be described in the system-instance aware perspective.\n\nThis property can only reflect the knowledge of the described system instance itself.\nOutside factors for availability can't need to be considered (e.g. network connectivity, middlewares).\n\nA disabled resource MAY skip describing its resource definitions.\n"
-    - name: 'minSystemVersion'
-      type: 'string'
-      description: "The resource has been introduced in the given [system version](../index.md#def-system-version).\nThis implies that the resource is only available if the system instance is of at least that system version."
+    - name: 'minSystemVersion_ID'
+      type: 'guid'
```
swennemers: @Fannon , is this change (string > guid) compatible ?

swennemers: otherwise I am fine with the PR

Fannon: Good catch. This is not a type change , but a new property added, which links the systemVersion to the describedSystemVersion as ID / association.

spec/v1/Document.schema.yaml
```diff
@@ -1029,6 +1029,15 @@ definitions:
           The resource has been introduced in the given [system version](../index.md#system-version).
           This implies that the resource is only available if the system instance is of at least that system version.
 
+          It MUST follow the [Semantic Versioning 2.0.0](https://semver.org/) standard.
+        x-introduced-in-version: "1.10.0"
+        x-association-target:
+          - "#/definitions/SystemVersion/version"
```
swennemers: I doubt this part

swennemers (approved):

static/spec-v1/interfaces/ums/MetadataType/apiresource.yaml
Fannon: Ok, this should not be an association anyway, as we're not pointing to a specific version but comparing with versions. I've removed x-assocation-targets from the source

spec/v1/Document.schema.yaml
Fannon: Yep, I removed it - you were right on this

PR #55, opened 2025-12-17
Fix broken anchor links and move Taxonomy to bottom

Fannon:
- Fixes most of the broken anchor tag links. 
- There are still warnings open, where docusaurus can't figure out the link target, because it's an ID of an `<a>` tag. This is the case for the generated interface documentation. 
- Moved Terminology Section to bottom, as it was getting too big
- Converted Terminology Section to headers instead of UL with inline #def- tags that docusaurus could not detect
- Where Terminology section was, added a mermaid diagram that shows most important concepts and links to the sections.

image


docs/spec-v1/index.md
```diff
@@ -407,6 +385,15 @@ Content-Type: application/json
 
 The response contains the requested resource and MAY include related ORD information that need to be updated together, e.g. when a data product is requested, it could also return its output ports API resources. Returning Tombstones is out of scope.
 
+#### Resource Definitions
+
+[ORD resources](#ord-resource) like APIs and Events reference [resource definitions](#resource-definition), which are machine-readable documents that describe the resource's interface in detail. These use industry-standard formats such as [OpenAPI](https://www.openapis.org/), [AsyncAPI](https://www.asyncapi.com/), JSON Schema, or WSDL.
```
maiargu, line 390:
```suggestion
[ORD resources](#ord-resource) like APIs and Events [resource definitions](#resource-definition), which are machine-readable documents that describe the resource's interface in detail. These use industry-standard formats such as [OpenAPI](https://www.openapis.org/), [AsyncAPI](https://www.asyncapi.com/), JSON Schema, or WSDL.
```

spec/v1/Document.schema.yaml
```diff
@@ -1635,8 +1636,10 @@ definitions:
           # Once there is a first standard, remove the sentence that only custom ones are supported in above description.
           - type: string
             pattern: ^([a-z0-9]+(?:[.][a-z0-9]+)*):(apiResource):([a-zA-Z0-9._\-]+):(v0|v[1-9][0-9]*)$
+            x-association-target:
```
maiargu, line 1639: what are the consequences in adding this?

Fannon: Only for documentation right now, but it seems like spec-toolkit is still missing this? In ORD Spec, we pointed out the association target explicitly as a link when this is provided. 

For UMS meta-model generation, it's also a hint, because there they need to know the association targets

PR #57, opened 2025-12-18
Documentation term use consistency

Fannon:
Fixing some inconsistent use of terms and upper/lowercasing things

Updated sequence diagram to be mermaid, as it also had some terms that needed updating


spec/v1/Document.schema.yaml
```diff
@@ -110,7 +110,7 @@ properties:
           The "system-independent" content MUST NOT be overridden via system-version or system-instance perspective metadata.
 
           Examples are: Vendors, products, globally aligned entity types, groups and group types (taxonomy), which are usually shared by multiple systems.
-    default: "system instance"
+    default: "system-instance"
```
Fannon, line 113: This was a bug, but I don't think it caused issues except creating inconsistent documentation. 

@maiargu - Maybe the spec-toolkit could not just validate the examples against the local JSON Schema, but also the default value? This would be a nice-to-have low prio feature.

Fannon: https://github.com/open-resource-discovery/spec-toolkit/issues/20

PR #59, opened 2025-12-18
Adding (AI) Agents as Top-Level Concept

HelgeCD:
This PR introduces **Agent** as a new top-level resource type in the Open Resource Discovery (ORD) protocol, enabling systematic discovery and modeling of AI agents across software ecosystems.

## Problem

Current ORD specification effectively handles APIs, events, and data products, but lacks a standardized way to describe AI agents beyond protocol-specific APIs. This creates challenges for:

- Discovering available agents in system landscapes
- Understanding agent capabilities and dependencies
- Modeling agents that exist internally without external invocation APIs

## Solution

Add a dedicated **Agent** resource type that:

- Inherits properties from ORD Capabilities (version, visibility, releaseStatus, etc.)
- Includes `exposedApiResources` to reference callable API interfaces
- Supports `integrationDependencies` to model system dependencies

## Key Benefits

- **Clear Separation**: Distinguishes between agent existence and API exposure
- **Non-callable Agents**: Models agents that perform internal functions without external access
- **Better Indexing**: Enables ORD aggregators to index agent metadata more effectively
- **Cross-Provider Alignment**: Provides consistent modeling approach across different providers
- **Extensibility**: Supports future AI ecosystem evolution (e.g. it does not rely on one specific API protocol)


spec/v1/Document.schema.yaml
```diff
@@ -2593,6 +2603,208 @@ definitions:
...
+        <<: *tags
+
+      labels:
+        <<: *labels
+
+      documentationLabels:
+        <<: *documentationLabels
+
+      systemInstanceAware:
+        <<: *systemInstanceAware
```
```suggestion
```
I don't think, we should have deprecated properties.

spec/v1/Document.schema.yaml
```diff
@@ -2593,6 +2603,208 @@ definitions:
...
+          maxLength: 255
+
+      changelogEntries:
+        <<: *changelogEntries
+
+      policyLevel:
+        <<: *policyLevel
+
+      customPolicyLevel:
+        <<: *customPolicyLevel
```
```suggestion
```

I don't think, we should have deprecated properties.

spec/v1/Document.schema.yaml
```diff
@@ -2593,6 +2603,208 @@ definitions:
...
+          - ["sap.odm:entityType:WorkforcePerson:v1"]
+
+      exposedApiResources:
+        type: array
+        description: |-
+          Optional list of API Resources that expose the functionality of the agent. Typically using the A2A protocol, but other protocols are possible as well.
+
+          MUST be a valid reference to an [API Resource](#api-resource) ORD ID.
+        items:
+          type: string
```
swennemers: Are you sure about the plain ORD ID array or better include objects [{ordId: "<ord id>"}] like we do for data product output ports?

I also assume the x-ums attribute is missing?! @Fannon

spec/v1/Document.schema.yaml
Fannon: You're right. Thanks for the catch

spec/v1/Document.schema.yaml
Fannon: I think the relationship type here is clear, so I don't really see how we'd need other properties on this. But I did fell into that trap before.. 

x-ums attributes, I can check then in January, as we'll schedule this for 1.14 release

spec/v1/Document.schema.yaml
Fannon: Agreed. It's a little bit inconsistent maybe - but then we force people who describe agents to switch to perspective features if we don't provide `systemInstanceAware` here.

docs/spec-v1/concepts/ai-agents-and-protocols.md
```diff
@@ -0,0 +1,50 @@
+---
+sidebar_position: 5
+description: AI Agents and related protocols in ORD.
+title: AI Agents and Protocols
+---
+
+# AI Agents and Protocols
```
Fannon, line 7: I like the title, it would make sense to also adjust the filename: `docs/spec-v1/concepts/ai-agents-and-protocols.md`

spec/v1/Document.schema.yaml
Fannon: I think we have all the x-ums properties defined, otherwise the generate script would fail.

Still undecided to go with `Array<ORD ID>` or `Array<{ordId: <ORD ID>}>`

Fannon: Decided to do `Array<{ordId: <ORD ID>}>`

swennemers (approved):

examples/documents/document-agents.json
```diff
@@ -0,0 +1,145 @@
...
+          "version": "1.0.3",
+          "url": "https://www.sap.com/foo/bar"
+        }
+      ],
+      "relatedEntityTypes": [
+        "sap.foo:entityType:DisputeCase:v1",
+        "sap.foo:entityType:DisputeResolution:v1",
+        "sap.odm:entityType:Dispute:v1"
+      ],
+      "exposedApiResources": ["sap.foo:apiResource:DisputeResolutionAgent:v1"],
```
swennemers:
```suggestion
      "exposedApiResources": ["ordId":"sap.foo:apiResource:DisputeResolutionAgent:v1"],
```

Fannon: Done, we made the change

docs/spec-v1/concepts/ai-agents-and-protocols.md
Fannon: done

spec/v1/Document.schema.yaml
Fannon: done

PR #63, opened 2026-01-05
fix: clarify documentation on apiProtocol: rest

maiargu:


spec/v1/Document.schema.yaml
```diff
@@ -1167,8 +1167,9 @@ definitions:
             description: |-
               Generic [REST](https://en.wikipedia.org/wiki/Representational_state_transfer) API.
               Please not that while OData is also a REST API, the most precise/opinionated `apiProtocol` MUST be chosen.
-              An API Resource definition of type `openapi-v3` (RECOMMENDED) or another appropriate option SHOULD be provided.
-              For each API Resource definition: `type` MUST ONLY be set to `openapi-v2`, `openapi-v3`, `openapi-v3.1+`, `raml-v1`, `sap-csn-interop-effective-v1` or `custom`.
+
+              If apiProtocol is `rest`, it means that at least one API Resource definition in the array `resourceDefinitions` MUST be provided that describes the API in a RESTful way.
+              At least one array item in `resourceDefinitions` MUST be of type `openapi-v2`, `openapi-v3`, `openapi-v3.1+` or `raml-v1`.
```
pavelkornev, line 1172:
```suggestion
              At least one array item in `resourceDefinitions` MUST be of type `openapi-v2`, `openapi-v3`, `openapi-v3.1+` or `raml-v1`.
              In addition, it MAY contain resource definitions of types `sap-csn-interop-effective-v1` and `custom`.
```

pavelkornev: We need explicitly list all possible options to avoid ambiguity.

Fannon: This is the general spec, I wonder if we can make such a MUST statement here - shouldn't this be handled on the policy level instead? 

We could make this a SHOULD statement if we want to be opinionated towards OpenAPI, which the community right now is. But then we should remove RAML as it didn't really won the adoption share.

```suggestion
              At least one array item in `resourceDefinitions` SHOULD be of type `openapi-v2`, `openapi-v3` or `openapi-v3.1+`.
              In addition, it MAY contain resource definitions of other types and `custom`.
              If the API Resource points to an `implementationStandard`, defining resource definitions is optional, as the resource definition is already described by the implementation standard itself.
```

pavelkornev: I agree on converting this to SHOULD in the general spec and enforcing to MUST in the SAP-specific policy level.

> In addition, it MAY contain resource definitions of other types and `custom`.

Shouldn't this statement apply to all API protocols, not just to `rest`?

pavelkornev: Wait, i find it inconsistent now with other API protocols. For each protocol we clearly list what allowed to be in the resource definition. If we say everything is allowed, this would mean that under `rest` it will be possible to expose `graphql-sdl`, `wsdl-v1`.. etc, which doesn't make any sense.

Fannon: Yes, you're right. This list was thought as listing those resource definitions that make sense in principle, and `custom` where we don't know and need to keep it open. 

Then we probably should still list CSN Interop because it's part of the standardized enum values and would work in principle.

swennemers: Seems as if I created the mess. As for the internal ORD resource for the data product extraction APIs, we have `rest`, but only CSN and no openapi (but with a special implementation standard. So, I would appreciate, to open up the spec to a SHOULD for openapi and have a specific exception for this implementation standard in the SAP and data product policies.  

Reg. @pavelkornev 's remark, I think, there is a difference between demanding the consumer to provide a specific resource definition type and having an allow list for resource definition types. We might want to separate the 2 aspects and document the allow list (MUST only be of type openapi, raml, csn, custom, SHOULD provide openapi - suggestion/policy level aspect)...

spec/v1/Document.schema.yaml
```diff
@@ -1167,8 +1167,9 @@ definitions:
             description: |-
               Generic [REST](https://en.wikipedia.org/wiki/Representational_state_transfer) API.
               Please not that while OData is also a REST API, the most precise/opinionated `apiProtocol` MUST be chosen.
```
```suggestion
              Please note that while OData is also a REST API, the most precise/opinionated `apiProtocol` MUST be chosen.
```

spec/v1/Document.schema.yaml
pavelkornev: > Seems as if I created the mess. As for the internal ORD resource for the data product extraction APIs, we have `rest`, but only CSN and no openapi (but with a special implementation standard. So, I would appreciate, to open up the spec to a SHOULD for openapi and have a specific exception for this implementation standard in the SAP and data product policies.

@swennemers, aren't those APIs of type `delta-sharing` with only CSN file as a resource definition file attached? REST APIs imply providing some metadata on how to do the actual data extraction (request) to an HTTP server.

swennemers: > @swennemers, aren't those APIs of type `delta-sharing` with only CSN file as a resource definition file attached? REST APIs imply providing some metadata on how to do the actual data extraction (request) to an HTTP server.

We are speaking on different use cases. Data Products have APIs type `delta-sharing` with CSN, correct. But the way how applications provide data for data products is with protocol `rest`, CSN and implementation standard data subscription API, see also https://github.com/cap-js/ord/issues/126

swennemers: I would appreciate if we can have the requirement for openapi resource definitions weakened when an implementation standard is provided. the implementation standard is defining the API contract already and typically externally, so we could reduce the requirement to recommendation (especially if tenant specific extensibility is available), but not mandatory.

Fannon: Agree, when there's an `implementationStandard` defined there is actually no strong need to provide a resource definition at all. After all, you're implementing a standardized resource definition contract and therefore it should be sufficient to point that out. 

I've updated the change proposal above

pavelkornev: I'm still unhappy with this statement:

>In addition, it MAY contain resource definitions of other types and `custom`.

"other types" is too vague and wide.

**Proposal:**

```yaml
- const: rest
  description: |-
    Generic [REST](https://en.wikipedia.org/wiki/Representational_state_transfer) API.
    Please not that while OData is also a REST API, the most precise/opinionated `apiProtocol` MUST be chosen.

    At least one array item in `resourceDefinitions` SHOULD be of type `openapi-v2`, `openapi-v3` or `openapi-v3.1+`.
    If the API Resource points to an `implementationStandard`, defining resource definitions is optional, as the resource definition is already described by the implementation standard itself.
    For each API Resource definition: `type` MUST ONLY be set to `openapi-v2`, `openapi-v3`, `openapi-v3.1+`, `sap-csn-interop-effective-v1` or `custom`.
```

pavelkornev: @Fannon speaking about the exception for `implementationStandard`. If it is set to a specific value, how is it possible to determine programmatically the actual REST operations of a rest API?

I would expect that implementation standards can be resolved somehow programmatically (namely getting OpenAPI file by its name)...

swennemers: > @Fannon speaking about the exception for `implementationStandard`. If it is set to a specific value, how is it possible to determine programmatically the actual REST operations of a rest API?
> 
> I would expect that implementation standards can be resolved somehow programmatically (namely getting OpenAPI file by its name)...

@pavelkornev , when using the implementationStandard, you typically are only interested that this implementationStandard is fulfilled as you know the conventions of it. In the example above with ds api, or also with exposing the ORD endpoint, you probably don't check the openapi files in generic manner (and it would be hard to provide them in every API implementing the standard...

I like your proposal from above

maiargu:
Closing this as PR discussions did not come to a final conclusion.

Fannon:
Let's keep it open as a reminder to clarify it. I have it on my TODO list, just other topics being more urgent.

PR #68, opened 2026-01-27
Discover namespaces via ORD Configuration

Fannon:
This is WIP and a proposal. 

With this we could use the ORD .well-known URI discovery to also find namespace JSON documents.
Those namespaces are what's used in the ORD IDs and other ORD related IDs.
The format for this is not yet publicly defined.


examples/documents/document-lifecycle.json
swennemers, line 1: @Fannon , should this be a namespaces.json? If not I am missing one.

Fannon: Oh, something got accidentally committed. We don't have the namespaces.json file yet, but we should add and link an example file here once the format is defined.

I'll clean this up. Thanks!

examples/configuration/configuration-1.json
```diff
@@ -35,6 +35,16 @@
         ]
       }
     ],
+    "namespaces": [
+      {
+        "url": "/namespaces.json",
```
Fannon, line 40: TODO: We should add this as an example in the /examples folder, too. 
We can do this along with then defining the JSON Schema for it.

PR #69, opened 2026-01-30
Extend documentation on AI Agent detail page

Fannon:


examples/documents/document-agents.json
```diff
@@ -94,7 +94,6 @@
       "exposedApiResources": [{"ordId": "sap.foo:apiResource:DisputeResolutionAgent:v1"}],
       "integrationDependencies": ["sap.foo:integrationDependency:DisputeCaseManagement:v1"],
       "labels": {
-        "llmModels": ["gpt-4", "claude-3.7-sonnet"],
```
Fannon, line 97: @HelgeCD I removed this because we already use the `partOfGroups`, which works better for this imho

HelgeCD: fair.
One way to model this is definitely better for the example, since it reduces the noise :)

PR #70, opened 2026-01-30
Use light-theme by default (fixed) but keep dark index page

Fannon:
The dark theme as default doesn't work so well for a spec page, where there is so much text and diagrams that presume lighter background. But I like how the dark index page looks. 

I tried to have the index page dark, light-theme is now the default for the rest of the page.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -104,7 +104,7 @@ const config = {
     /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
     ({
       colorMode: {
-        defaultMode: "dark",
+        defaultMode: "light",
         disableSwitch: false,
         respectPrefersColorScheme: false,
       },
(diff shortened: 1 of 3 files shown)


docusaurus.config.js
```diff
@@ -245,16 +245,10 @@ const config = {
...
+                  src="${baseUrl + "img/ord-footer-neonephos-light.svg"}"
                   alt="Neonephos Logo"
                   class="neonephos-logo neonephos-logo--dark"
                 />
-                <img
-                  src="${baseUrl + "img/ord-footer-neonephos-light.svg"}"
-                  alt=""
-                  aria-hidden="true"
-                  class="neonephos-logo neonephos-logo--light"
-                />
```
Fannon, line 276: Footer is now always light-theme

Fannon:
Closing as this has now been addressed with #128

PR #72, opened 2026-01-31
Use agolia-docsearch (docusaurus standard) and better link to NotebookLM

Fannon:
Add Algolia DocSearch v4 (default choice for public Docusaurus projects?) and better link to public ORD NotebookLM

### Search
- Replaced local search plugin with Algolia DocSearch v4
- Added Algolia site verification meta tag
- Removed `@easyops-cn/docusaurus-search-local` dependency

### Navigation
- Added "Ask AI" navbar pill linking to NotebookLM
- Added "Ask AI (NotebookLM)" to Help dropdown

### Documentation
- Added Quick Start section to overview page
- Made NotebookLM link more prominent with tip callout

@christianneu / @pavelkornev  - do you have any concerns using a public search service like Algolia? This is what docusaurus projects recommend by default. It would offer a better search experience than the fully local search, but obviously this would then include external requests. Not sure on the tradeoff here. 

Making the Notebook LM link more prominent is probably OK, because this is an external link and the user navigates to a separate service with their own terms and can then decide if they want to use it..


Fannon:
Won't do

PR #74, opened 2026-02-04
Add linting and formatting via Biome

Fannon:
Linting was lost when moving this to public repo, this adds it back but using Biome


Fannon:
outdated, it's been added in the meantime

PR #79, opened 2026-02-19
Support push-based transport mode

Fannon:
## Summary

This PR adds push transport as an optional ORD transport mode alongside pull.
It keeps the ORD Document interface transport-neutral and publishes heterogeneous resource definitions separately in their native media types.

- adds `POST /v1/documents` for identity-less ORD Document envelopes
- adds `PUT /v1/resource-definitions` for idempotent definition uploads identified by publisher, ORD ID, URL, and an unambiguous authorized publication context
- permits the aggregator to derive perspective, system version, and system instance from authenticated credential context, with request parameters used only where disambiguation is needed
- allows credential-specific restrictions without imposing a universal perspective restriction on categories such as customer-operated systems
- defines synchronous validation, resource-isolated outcomes, Problem Details errors, publisher handling of unsuccessful outcomes, and tombstone-based removal
- adds the normative ORD Aggregator Push API and updates the specification, mini-ADRs, and provider guidance

## Version-1 scope

Push support is optional.
An aggregator that supports it implements the two versioned endpoints and processes each request synchronously.
ORD documents and definitions may arrive independently and are revalidated as related content becomes available.
Providers should keep ORD documents within 2 MB.
Aggregators must support at least that size and may document a larger limit.

The ORD Document must still contain its perspective explicitly.
For resource-definition requests, `perspective`, `systemVersion`, and `systemInstanceId` may be omitted when the authenticated credential context determines exactly one applicable value.
Explicit request values never establish authority and must match the credential authorization and authoritative aggregator state.

Conditional definition updates with strong ETags and `Content-Digest` are optional optimizations.
Ordinary HTTP compression and transfer behavior are not ORD-specific capabilities.

## Open Questions / Discussion Points

- **Credential scope and request context:** The proposal allows an aggregator to bind a credential to perspectives, system versions, or system instances and infer context where this produces exactly one result. Should version 1 recommend narrowly scoped credentials more strongly, or is the current flexibility needed for multi-tenant publishers that serve several contexts through one implementation?
- **Publisher and perspective restrictions:** The proposal deliberately does not state that a category such as a customer-operated system may publish only `system-instance` metadata. Such a rule would incorrectly generalize current implementation constraints and exclude platforms that publish customer-created content through a valid static perspective. Should any interoperable restriction be added beyond credential-specific authorization policy?
- **Error handling:** A successful HTTP response can contain rejected or stale item outcomes. Publishers must inspect per-item results and retain or expose diagnostics, while the route by which an operator or developer receives those diagnostics remains implementation-specific. Is this sufficient for automated customer-system publishers?
- **Removing content:** Each push contains updates, not a complete replacement of the current content. Omitting an ORD entry therefore does not remove it. The ORD Provider must publish a tombstone for every removed entry. If it misses a tombstone, the ORD Aggregator keeps serving stale content. Is this reliable enough for providers, or should the push API support publishing a complete set of content and let the aggregator determine what was removed?
- **Scope of a complete replacement:** Several ORD Providers may publish content for the same system type, system version, system instance, or perspective. One provider must not remove content published by another provider. If the API supports complete replacement, how does the aggregator identify the exact set of content that the authenticated provider owns and is allowed to replace?
- **Direct uploads or transactional submissions:** The proposed version-1 API processes each document or resource definition independently and returns the result synchronously. A submission resource could stage several uploads, validate them together, and publish them with one commit. Should submissions and commit semantics be part of version 1, or is it sufficient to add them later?

## Future extensions

Version 1 does not define endpoint discovery, batching, submissions, asynchronous operations, or commit semantics.
A future submission resource could group uploads and expose asynchronous validation status.


docs/spec-v1/index.md
```diff
@@ -193,18 +193,193 @@ Manual import of the [ORD document](#ord-document) as a JSON file into an intere
...
+
+- Static metadata that is known at design-time or deploy-time
+- CI/CD pipeline integration where metadata is pushed as part of the build/deployment process
+- Providers that cannot or prefer not to host a runtime ORD Provider API
+
+##### Push Transport - Pros
+
+- No need to implement and host an ORD Provider API (simpler provider implementation)
+- Can be integrated into CI/CD pipelines (design-time or deploy-time)
+- Changes are pushed immediately when they occur (no polling delay)
```
swennemers:
```suggestion
- Configuration or extensibility changes are pushed immediately when they occur (no polling delay)
```

docs/spec-v1/index.md
```diff
@@ -193,18 +193,193 @@ Manual import of the [ORD document](#ord-document) as a JSON file into an intere
...
+
+An [ORD aggregator](#ord-aggregator) that supports push transport MUST provide a dedicated push API endpoint for receiving ORD documents.
+
+<!-- The aggregator MAY advertise its push API capabilities via its own ORD configuration endpoint at `/.well-known/open-resource-discovery`.
+The aggregator configuration MAY include:
+- The URL of the push API endpoint
+- Supported access strategies for authenticating providers
+- Description and documentation links for onboarding
+
+> For a more detailed draft proposal of the aggregator configuration schema, see [Aggregator Configuration (Draft)](./concepts/aggregator-configuration.md).
```
swennemers: @Fannon this concept document doesn't exist in this PR?

Fannon: Ah yes, I deleted it to keep the PR more lean. Technically it's not necessary as each aggregator can just document its behavior. It's still in this commit just in case : https://github.com/open-resource-discovery/specification/pull/79/commits/68f7c5a0fc27378e6a3a89ebf6113c8da5ff09bc

```suggestion
```

docs/spec-extensions/access-strategies/embedded.md
```diff
@@ -0,0 +1,95 @@
...
+
+> **Status**: Draft Proposal (WIP)
+>
+> This access strategy is part of the [Push Transport](../../spec-v1/index.md#push-transport) proposal.
+
+## Description
+
+The `embedded` access strategy indicates that the resource definition content is provided inline within the ORD document itself, rather than being fetched from an external URL.
+
+This is specifically designed for [push transport](../../spec-v1/index.md#push-transport) scenarios where:
```
SvetlinZarev: Does "designed for" mean "limited to"? Embedded resources seem equally useful in pull mode — a provider might want to serve a self-contained ORD document without requiring the aggregator to make additional HTTP requests for resource definitions. IMO push vs. pull is purely a transport concern and shouldn't dictate what the document can contain.

docs/spec-extensions/access-strategies/embedded.md
```diff
@@ -0,0 +1,95 @@
...
+
+# Embedded Access Strategy
+
+> **Status**: Draft Proposal (WIP)
+>
+> This access strategy is part of the [Push Transport](../../spec-v1/index.md#push-transport) proposal.
+
+## Description
+
+The `embedded` access strategy indicates that the resource definition content is provided inline within the ORD document itself, rather than being fetched from an external URL.
```
SvetlinZarev: Why inline instead of a ZIP file? 

Embedded resources could easily push the document past the 2MB limit. Also it may greatly increase the memory requirements for parsing the document.

Example:
```
/
- MANIFEST.MF
- resource-1.ord.json
- resource-M.ord.json
- api-123.openapi.json
- api-XYZ.openapi.json
```

With a zip file you can push an arbitrary amount of resource definitions (and even ORD documents if it makes sense).

All files should be listed in the manifest. It acts as the document served at the "well-known" endpoint. 

Optionally it can contain signatures, hashes, etc.

docs/spec-extensions/access-strategies/embedded.md
```diff
@@ -0,0 +1,95 @@
...
+- Parse the string content according to the `mediaType` specified
+
+## When to Use
+
+Use the `embedded` access strategy when:
+
+- Pushing ORD documents to an aggregator (push transport)
+- Integrating ORD publishing into CI/CD pipelines
+- The provider cannot or does not want to host resource definitions at accessible URLs
+- You want to ensure atomic updates of metadata and definitions together
```
SvetlinZarev: Should atomic "all-or-nothing" semantics be a hard requirement here? If a push partially fails (e.g. some resource definitions are invalid), should the aggregator reject the entire payload or accept the valid parts? 

This needs an explicit statement in the spec — partial updates could leave an aggregator in an inconsistent state, but strict atomicity may be impractical for large payloads.

docs/spec-v1/index.md
```diff
@@ -193,18 +193,192 @@ Manual import of the [ORD document](#ord-document) as a JSON file into an intere
...
+- Centralized approach (aggregator must be available to receive pushes)
+
+##### Push Transport Implementation
+
+###### ORD Push Document
+
+For push transport, the standard [ORD document](#ord-document) format is used with one addition: a `definitions` property that allows inline [resource definitions](#resource-definition).
+
+When using pull transport, resource definitions are referenced via URLs and fetched separately by the aggregator.
+In push transport, these definitions can be provided inline within the ORD document itself using the `definitions` property.
```
SvetlinZarev: There's no need to limit this to push transport. Most aggregator implementations will have a single document-handling pipeline regardless of how the document arrived — the transport layer should be transparent to the processing logic. Artificially restricting embedded resources to push mode adds complexity for no benefit.

docs/spec-v1/index.md
```diff
@@ -193,18 +193,192 @@ Manual import of the [ORD document](#ord-document) as a JSON file into an intere
 - The ORD document alone is sufficient for this type of consumption
 - All URLs in the document MUST be resolvable (e.g. through `baseUrl` or full URLs)
 
-#### Push Transport
+### Push Transport
```
SvetlinZarev, line 182: This proposal feels underspecified for enabling reusable, interoperable aggregators. For anyone to build a generic aggregator that works across providers, there needs to be a standard push protocol — not just a payload shape. Concretely, that means defining:
  
  - A standard push endpoint (or a standard discovery mechanism for it, e.g. a well-known)
  - Transport mode — multipart, streaming, ZIP, plain JSON body? This matters for the 2MB-per-document constraint: embedding resource definitions inline as strings inside a JSON payload  doesn't scale, and the choice of transport directly determines whether large payloads are even feasible
  - Sync vs. async communication — I'd strongly advocate for async here. A fire-and-forget push with a callback/webhook/polling for validation results would allow providers to push arbitrarily large amounts of data without blocking the CI/CD pipeline on aggregator processing time. Sync push only works for small, fast payloads; async is the only model that scales

  Without these being standardized, every aggregator will implement its own variant and the push mechanism won't be reusable — which defeats the purpose.

Fannon: Yes, that's on point. Right now it's underdefined, reason is that we would need to build / prototype this along with defining here in the spec and that hasn't happened yet. I agree that those questions need to be clarified. 

On the async side, it comes with some drawbacks like not getting errors / problems quick enough. But since the aggregator needs a bit time to validate, maybe it's more realistic to batch more and then also receive validation asynchronously later. It would be a more complicated pattern than just sync push with direct error feedback.

docs/spec-v1/index.md
Fannon: Agreed, it would also make sense on pull. 
Where I'm not sure: It would be a new feature not every consumer would support, so it's an incompatible change if we backport it to pull transport. So it feels a bit dangerous to introduce it to pull without having a major release (which I'd like to avoid). 

Conceptually, it would of course work for both transport modes.

docs/spec-extensions/access-strategies/embedded.md
Fannon: Good point, so far we left it to the aggregator as an implementation detail. I think there's some "oppinion" in how strong this is fulfilled. 

I'll keep it in mind, but defining this would be probably a general enhancement, not just for the push based appraoch.

docs/spec-extensions/access-strategies/embedded.md
Fannon: Yes, I also thought about it. It would be a different approach, then we don't need embedding. Instead we use relative file links. 

Those ZIP files could get quite big though and I'm not sure how ideal it is with regards to HTTP protocol which already does compression as part of the transport. 

I see another alternative: POST/PUT all the files as part of a transaction, which then can be committed via separate action on the API. Then there can be a GET on the transaction to also receive feedback. 
This would be closer to HTTP conventions and reuse of how the protocol works, also would be more async. It requires some transaction management on aggregator side, though.

docs/spec-extensions/access-strategies/embedded.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Description
+
+The `embedded` access strategy indicates that the resource definition content is provided inline within the ORD document itself, rather than being fetched from an external URL.
+
+This is specifically designed for [push transport](../../spec-v1/index.md#push-transport) scenarios where:
+
+- The ORD provider pushes an ORD document together with the resource definitions referenced by its entries
+- The aggregator does not need to make additional requests to fetch resource definitions
+- Each pushed document is self-contained with respect to its own referenced definitions
```
pavelkornev: This might be a problem. What does it mean `self-contained`? For example, ABAP System has 500 ORD Documents to describe 100K APIs each API Resource reference API Package which is defined once in a separate ORD Document. Does it mean, that provider would need to duplicate some definitions to make ORD Documents `self-contained`?

docs/spec-extensions/access-strategies/embedded.md
```diff
@@ -0,0 +1,97 @@
+---
+title: Embedded Access Strategy
```
pavelkornev: Why do we need this access strategy at all? `accessStrategy` fields are optional. In case of `push`, it can be simply omitted for simplicity. And we can have a rule in the AMV, if there is no `accessStrategy` defined, then `url` properties must contain local paths.

docs/spec-v1/index.md
```diff
@@ -178,31 +177,189 @@ This is implemented by providing an [ORD Provider API](#ord-provider-api).
...
+For push transport, the standard [ORD document](#ord-document) format is used with one addition: a `definitions` property that allows inline [resource definitions](#resource-definition).
+
+When using pull transport, resource definitions are referenced via URLs and fetched separately by the aggregator.
+In push transport, these definitions can be provided inline within the ORD document itself using the `definitions` property.
+
+When embedding definitions, the resource definition's access strategy MUST be set to [`embedded`](../spec-extensions/access-strategies/embedded.md) to explicitly indicate that the content is inline rather than fetched from a URL.
+
+The `definitions` property is a dictionary where:
+- The **key** is the URL path (as referenced by resources via `resourceDefinitions[].url`)
+- The **value** is the raw content of the resource definition as a **string**
```
pavelkornev: Do you really mean embedding resource definition inside the ORD Documents? Like Base64?

docs/spec-v1/index.md
```diff
@@ -178,31 +177,189 @@ This is implemented by providing an [ORD Provider API](#ord-provider-api).
...
+          "accessStrategies": [
+            { "type": "embedded" }
+          ]
+        }
+      ]
+    }
+  ],
+  "definitions": {
+    "/api/my-api/openapi.json": "{\"openapi\":\"3.0.0\",\"info\":{\"title\":\"My API\",\"version\":\"1.0.0\"}}"
+  }
```
pavelkornev: This will elevate the size of ORD Documents unpredictable which could cause also issues on the validator size. Aggregator would need to remove this before sending such a file to the validator.

docs/spec-v1/index.md
```diff
@@ -178,31 +177,189 @@ This is implemented by providing an [ORD Provider API](#ord-provider-api).
...
+See the [static baseline](/spec-v1/examples/document-push-transport) and [system-instance delta](/spec-v1/examples/document-push-transport-delta) push examples.
+
+##### ORD Aggregator Push API
+
+An [ORD aggregator](#ord-aggregator) that supports push transport MUST provide a dedicated push API endpoint for receiving ORD documents.
+Discovery and configuration of that endpoint are outside the scope of this proposal.
+
+##### Push API Contract
+
+The push API endpoint MUST:
```
pavelkornev: What about large content? Aggregator needs to accept multiple zips with chunks, unzip it, then process as a whole thing.

Fannon:
### Meeting Notes:

- We decided not to to embedding into ORD documents, instead we favor separate requests to push the resource definitions. The ORD Document format may not have to change at all then. 
- The "url" could become the ID of the definition that is pushed, but additional context is needed, e.g. which ORD ID / tenant this is pused to. The "url" is only unique within a tenant ORD resource.
- How the push API contract looks like is yet to be defined
- zipping would be handled by the HTTP protocol / infrastructure transparently, we anyway only push one document / file per request. If we want to batch, a possible optimization would be multipart/related or streaming ? This has to be an optional optimization though. 
- To be clarified how security / auth / identy works, this will be specific to the aggregator to some degree. Unclear how much the ORD spec can define here generically. 
- Access Strategy is not applicable in push mode, can be ommitted

Fannon:
I updated the proposal based on Thomas's feedback and merged the latest `main`.

The main clarification is that push does not impose a universal perspective restriction based on publisher category.
A customer-operated system may commonly be limited to `system-instance` in a particular onboarding setup, but making that a general ORD rule would exclude platforms that validly publish customer-created content through a static perspective.
An aggregator can instead restrict each credential to particular perspectives, system versions, or system instances.

For resource-definition uploads, `ordId` and `url` remain mandatory.
The aggregator should derive `perspective`, `systemVersion`, and especially `systemInstanceId` from authenticated credential context when it can resolve exactly one authorized value.
The request supplies context parameters only when disambiguation is needed, and supplied values must match authoritative server state rather than establish authority themselves.

The response contract now also states that publishers must inspect per-item outcomes even after `200 OK`, treat `rejected` and `stale` outcomes as unsuccessful updates, retain or expose diagnostics, retry transient failures with backoff, and avoid retrying unchanged validation or authorization failures.
How those diagnostics reach a developer or operator remains implementation-specific.

These changes are reflected in the specification and OpenAPI contract and in mini-ADRs 003, 004, and 005 so the rationale and alternatives remain visible for discussion.

Fannon:
Thanks, this identifies a useful distinction.

`POST /v1/documents` does not create a document resource.
It updates the contained ORD items by their own identities, so updating an item does not require a preceding GET or DELETE of an earlier document.

The ABAP case reveals a separate missing capability: scoped current-state replacement.
A provider that knows its complete current state, but not everything it published historically, should be able to declare that set and let the aggregator derive removals.

The replacement boundary must not be the ORD Document or merely an ORD ID namespace.
It should be the contributions attributed by the aggregator to one authenticated publisher in exactly one publication context.
Namespace authorization remains a separate check.
This preserves content from other publishers or delegators, for example BDC-contributed content associated with an S/4 system, even if the same system context or an overlapping authorized namespace is involved.
Credential rotation also must not change that stable publisher attribution.

A complete S/4 state can span many ORD Documents, so a single `PUT /v1/documents` cannot express this replacement safely.
One possible extension to the proposed API would let multiple existing `POST /v1/documents` requests carry one publication-cycle identifier.
After all documents were processed successfully, the provider would send an explicit completion signal.
The aggregator could then remove contributions in the same replacement scope that were not observed during that cycle.
This reuses the proposed per-item ingestion model, but it still requires cycle identity, completion, failure, concurrency, retry, and expiry semantics that version 1 does not currently define.

The transactional submission style handles this requirement more naturally.
It can stage all documents and definitions, validate the complete set, and retain the previous published state until an atomic commit succeeds.
The submission endpoints sketched in ADR 005 remain a non-normative design variant rather than a change to the proposed version-1 API, but the replacement requirement is an important argument in favor of that style.

A projected list of ORD IDs is more naturally an Aggregator Discovery API capability and can be useful for reconciliation.
It is not equivalent to replacement and must not expose a merged set as a deletion inventory.

adrs/005-use-simple-artifact-endpoints-with-optional-optimizations.md
```diff
@@ -0,0 +1,133 @@
...
+Not chosen for version 1 (see [ADR 002](./002-use-eventually-consistent-resource-isolated-ingestion.md), where adopting it from the start is weighed as an alternative baseline).
+The following endpoint sketch records one possible design discussed during review; it is illustrative and is not the standardized implementation.
+A submission would be a staging area: uploads made against it would not be discoverable until commit, and commit would publish them atomically.
+
+```text
+POST   /v1/submissions{?scope}         open a submission → { submissionId }
+POST   /v1/submissions/{id}/documents            stage a document (same body as /v1/documents)
+PUT    /v1/submissions/{id}/resource-definitions stage a definition (same params as /v1/resource-definitions)
+POST   /v1/submissions/{id}/commit{?mode}         publish atomically; mode=replace treats the staged set as complete
+DELETE /v1/submissions/{id}                       abandon; discard all staged uploads
```
Fannon, line 123: Auto-discard if submission hasn't been committed within a certain timeframe. 

Need endpoint to get state of submission? Endpoint to get all errors/info/warn of that submission.

adrs/005-use-simple-artifact-endpoints-with-optional-optimizations.md
```diff
@@ -0,0 +1,133 @@
...
+
+Not chosen for version 1 (see [ADR 002](./002-use-eventually-consistent-resource-isolated-ingestion.md), where adopting it from the start is weighed as an alternative baseline).
+The following endpoint sketch records one possible design discussed during review; it is illustrative and is not the standardized implementation.
+A submission would be a staging area: uploads made against it would not be discoverable until commit, and commit would publish them atomically.
+
+```text
+POST   /v1/submissions{?scope}         open a submission → { submissionId }
+POST   /v1/submissions/{id}/documents            stage a document (same body as /v1/documents)
+PUT    /v1/submissions/{id}/resource-definitions stage a definition (same params as /v1/resource-definitions)
+POST   /v1/submissions/{id}/commit{?mode}         publish atomically; mode=replace treats the staged set as complete
```
Fannon, line 122: Add mode=strict

Fannon: Or make strict mode default? This assumes no error must happen in transaction, otherwise it's not accepted. ORD Provider has to fix, can resubmit the document / resource definition into same submission (?)

PR #80, opened 2026-02-26
Add i18n Support

HelgeCD:
_Still a WIP branch._ 

Before starting any actual implementation I created an ADR based on some SAP internal discussions. Goal is to be able to gather some more feedback before actually starting defining the specification.

Adding the ADR to the repo was aligned with @Fannon. The goal is to be able to understand the design decisions long term and outside the scope of a single PR. Not sure if the current structure is ideal, but I think it is a good starting point.

**Next Steps**: Gather Feedback (~2 weeks) and refine specification afterwards.

### DIFF
--- a/adr/0001-internationalization-support.md
+++ b/adr/0001-internationalization-support.md
@@ -1,0 +1,268 @@
+# ADR-0001: Internationalization Support
+
+## Table of Contents
+
+- [Status](#status)
+- [Context](#context)
+- [Decision](#decision)
+  - [Sub-Options Considered for Approach 1](#sub-options-considered-for-approach-1)
+  - [Core Design Principles](#core-design-principles)
+  - [Schema Structure](#schema-structure)
+  - [ORD Configuration Integration](#ord-configuration-integration)
+  - [Translatable Properties](#translatable-properties)
+  - [Handling Complex Objects (Links)](#handling-complex-objects-links)
+- [Consequences](#consequences)
+  - [Positive](#positive)
+  - [Negative](#negative)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Alternatives Considered](#alternatives-considered)
+  - [Approach 2: Embed into ORD Document](#approach-2-embed-into-ord-document)
+  - [Approach 3: Separate Translation File (gettext-style)](#approach-3-separate-translation-file-gettext-style)
+- [Future Considerations](#future-considerations)
+  - [Definition File Translation](#definition-file-translation)
+  - [Adoption Strategy](#adoption-strategy)
+- [References](#references)
+
+## Status
+
+Proposed
+
+## Context
+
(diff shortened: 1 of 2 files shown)


adr/0001-internationalization-support.md
```diff
@@ -0,0 +1,268 @@
...
+**Resource-Specific Properties:**
+- API/Event Resources: `customImplementationStandardDescription` - Helps technical decision-makers understand API contracts
+- Links: `title`, `description` - User-facing when displayed in catalogs (requires array index-based identification)
+- Vendor: `title` - Displayed in catalogs (though often company names that may not need translation)
+
+#### Low Priority (Technical/Support Audience)
+
+**Support-Oriented Properties:**
+- Package: `supportInfo` - Primarily for support teams, not end-users
+- Extensible objects: `extensible.description` - Technical information for developers
```
swennemers, line 133: Extensibility should in general move into the end user/key user = business user responsibility and this information is shown in BAH/catalogs, therefore, we should enable translation for this property as well. The information here is necessary for key users to understand, what they need to select in their translated key user extensibility tools.

adr/0001-internationalization-support.md
```diff
@@ -0,0 +1,268 @@
...
+## Decision
+
+We will implement **Approach 1: Dedicated ORD i18n Format** with a **fixed property key set** rather than generic JSON paths.
+
+### Sub-Options Considered for Approach 1
+
+**Option 1A: Generic JSON Paths**
+- Use JSON Pointer syntax (e.g., `"$.title"`, `"$.shortDescription"`) to target translatable properties
+- More flexible but risks targeting technical fields like `$.version` or `$.ordId`
+- Fragile to schema changes and requires deep ORD structure knowledge
```
swennemers, line 51: can we move this to the rejected options?

adr/0001-internationalization-support.md
```diff
@@ -0,0 +1,268 @@
...
+    }
+  }
+}
+```
+
+**Limitation**: This approach is fragile when link arrays are reordered. The JSON path approach (Option 1A) would handle complex objects more elegantly, but the overall benefits of type safety and preventing accidental translation of technical fields still favor the fixed property approach.
+
+**Recommendation**: For the initial implementation, focus on the core translatable properties and defer complex object translation to a future version.
+
+## Consequences
```
swennemers, line 159: can we specify the impact and expectations on aggregators and consumers?
Especially:
* Are aggregators expected to return the i18n info embedded into the ord resources to consumers?
* Are consumers able to query with a target language and get texts in the target language with a language fallback? (like browsers/UIs do it)

swennemers (commented):
@HelgeCD @Fannon , I reviewed the proposal, I am fine with the direction. Please see my comments for details.

adr/0001-internationalization-support.md
swennemers: do we have aggregements on implementing this with our main aggregators?

PR #81, opened 2026-02-26
Hide properties and entities that have been deprecated for 2-3 releases

Fannon:
Goal is to make the human readable documentation less daunting by hiding properties and entities that have been deprecated for a longer time already. They are just hidden from the documentation and still part of the JSON Schema / Interfaces. So this is not a breaking change, just cosmetic cleanup. 

Alternative idea: Generate the documentation twice, so there is also a version where everything can be seen. But by default show the cleaned up version.

### Changed

- Hidden deprecated properties from documentation using `x-hide: true`:
  - `policyLevel` and `customPolicyLevel` (deprecated since v1.9.9) - use `policyLevels` instead
  - `entityTypeMappings` (deprecated since v1.11.0) - use `exposedEntityTypes` instead
  - `systemInstanceAware` (deprecated since v1.12.0) - use `perspective` instead
  - Related definitions: `EntityTypeMapping`, `ApiModelSelectorOData`, `ApiModelSelectorJsonPointer`, `EntityTypeOrdIdTarget`, `EntityTypeCorrelationIdTarget`
- Updated documentation to reference current concepts instead of deprecated ones:
  - Changed `policyLevel`/`customPolicyLevel` references to `policyLevels` in spec index
  - Changed `systemInstanceAware` references to `perspective` in access strategy docs and FAQ
- Removed deprecated property usage from example files (`systemInstanceAware`, `policyLevel`, `customPolicyLevel`)


spec/v1/Document.schema.yaml
```diff
@@ -1450,6 +1450,7 @@ definitions:
 
       entityTypeMappings: &entityTypeMappings
         type: array
+        x-hide: true
```
swennemers, line 1506: I am not sure, if we can already hide it already. BAH is not yet supporting exposedEntityTypes, S/4 is still exposing it...

swennemers (commented):
in general +1, but I have doubts on hiding entityTypeMapping yet

spec/v1/Document.schema.yaml
Fannon: Yes, but we would just hide it from documentation. For new adopters, we don't want them to implement / see this. Also it looks like there is now no consumer for it, except BAH using it to create the links (which would be simpler with `exposedEntityTypes`

Hiding this would greatly help cleaning things up, as it also removes some objects

swennemers: I understand, but the support for exposedEntityTypes is not given and we still want the links and we haven't updated all guidelines to use exposedEntityTypes, do we? So documentation is still needed until one can really use exposedEntityTypes.

Fannon: Then let's push BAH to support it. And the guidelines should be updated already I guess.. we'd need to check

Fannon: @swennemers do you have strong objectsions? I would prefer to do this sooner than later, as it cleans up and reduces chances that deprecated features are used for new developments. 

@maiargu - how easy would it be to support in spec-toolkit that we generate the markdown documentation twice, once with and once without hidden content? Is this something we should do from the beginning?

CHANGELOG.md
```diff
@@ -10,6 +10,18 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ## [unreleased]
 
+### Changed
+
+- Hidden deprecated properties from documentation using `x-hide: true`:
```
maiargu, line 15: I think changes done to the UI by appling `x-hide` don't need to be added to the changelog file, the exposed npm package JSON schema did not change at all for consumers

Fannon: I'd still prefer to keep it in, because many read the documentation as a page and the changelog might at least give a hint why some things have now disappeared. Better over-communicate here :) But yes, for JSON Schema / NPM consumption, it's not changing anything

PR #82, opened 2026-03-02
PoC: Add ORD Overlays 

Fannon:
This is just a Draft / PoC: We could use the ORD Spec to define model extensions (UMS Metadata Type extensions) that are related to ORD concepts (link via ORD IDs). This would leverage the existing ORD documentation and spec generation pipelines.

Still to be decided if this is a a good idea and what model extensions we should define along with ORD spec. Each model extension would have its own version / lifecycle, potentially ownership.


swennemers:
I am fine with adding model extensions to the spec, however, the concrete overlay extension is too limited to give substantial feedback. Except that I suggest calling it Resource Definition Overlays, to indicate the difference between API Resources and Resource Definitions, as terms the ORD spec is using, API Metadata is unspecific in regards to which of these files should be overlayed.

Fannon:
> I am fine with adding model extensions to the spec, however, the concrete overlay extension is too limited to give substantial feedback. Except that I suggest calling it Resource Definition Overlays, to indicate the difference between API Resources and Resource Definitions, as terms the ORD spec is using, API Metadata is unspecific in regards to which of these files should be overlayed.

Yes, it's actually empty right now :) Thanks for the naming feedback, definitely will rename it as you suggested

PR #85, opened 2026-03-05
Clarify allowed values for `type` field 

Fannon:
### Changed

- Clarification: The `type` field on `ApiResourceDefinition` and `EventResourceDefinition` accepts not only the predefined enum values and `custom`, but also any valid [Specification ID](./docs/spec-v1/index.md#specification-id) as an extension point. All `apiProtocol` descriptions that listed allowed `type` values now consistently mention this option alongside `custom`.


spec/v1/Document.schema.yaml
```diff
@@ -1213,40 +1213,40 @@ definitions:
             description: |-
               [OData Version 2.0](https://www.odata.org/documentation/odata-version-2-0/) API.
               An API Resource definition of type `edmx` MUST be provided.
-              For each API Resource definition: `type` MUST ONLY be set to `edmx`, `csdl-json`, `openapi-v2`, `openapi-v3`, `openapi-v3.1+`, `sap-csn-interop-effective-v1` or `custom`.
+              For each API Resource definition: `type` MUST ONLY be set to `edmx`, `csdl-json`, `openapi-v2`, `openapi-v3`, `openapi-v3.1+`, `sap-csn-interop-effective-v1`, `custom` or any valid [Specification ID](../index.md#specification-id).
```
Fannon, line 1216: I wonder if it would be enough to stick to SHOULD. 

We already previously had the problem that we added a new resource definition format that was now also a valid option and then we had to first update validation in order to be able to adopt it. Having this as an error is not "forward compatible". FYI, @pavelkornev , @swennemers 

```suggestion
              For each API Resource definition: `type` SHOULD ONLY be set to `edmx`, `csdl-json`, `openapi-v2`, `openapi-v3`, `openapi-v3.1+`, `sap-csn-interop-effective-v1`, `custom` or any valid [Specification ID](../index.md#specification-id).
```

swennemers: @pavelkornev @maiargu had good arguments that validation with SHOULD is hard and consumers like BAH expect a limited set of supported definition types. => @Fannon, However you describe it here, the SAP internal policy levels will disallow it anyhow.

I appreciate to have the spec ID in addition, as we might inlcude the AMS policies as additional definition type.

Fannon: Let's add AMS policies in separate PR, maybe they're just a Specification ID? Then we don't add more SAP owned formats to the public spec.

Fannon: Decided not to do it. With that I'll just merge the PR as it is.

PR #86, opened 2026-03-05
ORD Overlay Specification (alpha)

Fannon:
The **ORD Overlay** is an optional ORD model extension that allows patching various referenced resource definition files (e.g. OpenAPI, AsyncAPI, OData CSDL, MCP/A2A Agent Cards) without modifying the original source files. Fore the future, it's planned to also support patching ORD level information as well, hence the opinionated name. 

### What to Review

This PR contains a lot of file changes, but some of them are just generated files (ignore in review) and quite a few are for the merge-overlay script that we should move to a separate repository / NPM project later anyway. 

For easier reviewing, concentrate on: 
- https://open-resource-discovery.github.io/pr-preview/specification/pr-86/spec-extensions/models/OrdOverlay (this is the end-user documentation)
- https://github.com/open-resource-discovery/specification/tree/ord-overlays/examples/overlay (overlay example files)

### Added

- Added **ORD Overlay** as an alpha spec extension model (`ord:overlay:v1`)
  - Overlays allow patching ORD resource metadata and referenced definition files (OpenAPI, AsyncAPI, OData CSDL, MCP/A2A Agent Cards) without modifying the originals.
  - Overlays can be distributed via the ORD Configuration Endpoint (`openResourceDiscoveryV1.overlays`) or attached directly to an API/Event resource as a `resourceDefinitions` entry.
  - Patches use concept-level selectors (`ordId`, `operation`, `entityType`, `propertyType`) or a generic `jsonPath` fallback, with actions `merge`, `update`, `append`, and `remove`.
  - Optional `target` object narrows a patch to a specific definition file or format (e.g. `definitionType: openapi-v3`).
  - Optional top-level fields (`describedSystemType`, `describedSystemVersion`, `describedSystemInstance`, `visibility`) scope the overlay to a particular system context.


docs/spec-extensions/models/OrdOverlay.md
```diff
@@ -0,0 +1,460 @@
...
+An ORD Aggregator SHOULD enforce that overlay sources are permitted to patch the target metadata.
+Without such enforcement, consumers could be exposed to unauthorized metadata changes through overlay processing.
+
+## Overlay Document Metadata
+
+Optional top-level fields scope an overlay to a specific system context:
+
+- [`perspective`](#overlay-perspective) — declares whether the overlay applies at system-type, system-version, or system-instance scope.
+- [`describedSystemType`](#overlay-system-type), [`describedSystemVersion`](#overlay-system-version), [`describedSystemInstance`](#overlay-system-instance) — narrow the overlay to a particular system type, version, or instance.
+- [`visibility`](#overlay-visibility) — controls who can discover this overlay document (`public`, `internal`, `private`).
```
swennemers:
```suggestion
- [`visibility`](#ord-overlay_visibility) — controls who can discover this overlay document (`public`, `internal`, `private`).
```

swennemers: fixing link

docs/spec-extensions/models/OrdOverlay.md
```diff
@@ -0,0 +1,460 @@
...
...ntirely with `data`.<br/>- `append`:<br/>  - append string `data` to the selected string value.<br/>  - only valid when the selected element is a text/string field.<br/>  - useful to extend existing descriptions without replacing them.<br/>- `merge`:<br/>  - objects are deep-merged recursively.<br/>  - scalar values are overwritten by the value from `data`.<br/>  - arrays are appended (new array items are added after existing items).<br/>  - existing object properties not mentioned in `data` are preserved.<br/><br/>  To fully replace an array, use two ordered patches:<br/>  1. `remove` the array at the selected location.<br/>  2. `merge` the new array content.<br/>- `remove`:<br/>  - without `data`: remove the selected element from the document entirely.<br/>  - with `data`: remove fields that are set to `null`.<br/>    This applies recursively, so nested `null` values remove nested fields as well<br/>    (JSON Merge Patch-style delete semantics).<br/><br/>Example for nested removal:<br/>`data: { "foo": { "bar": null } }` removes `foo.bar` inside the selected element.<hr/>**Allowed Values**: <ul><li>`"update"`</li><li>`"append"`</li><li>`"merge"`</li><li>`"remove"`</li></ul></div>|
```
swennemers: 2. level indentation is wrong

swennemers: switch order of remove and merge, because merge is referencing the definition of remove

swennemers (commented):
@Fannon , I like the proposal, comments are already on the UA level

examples/overlay/edmx-business-partner.overlay.json
```diff
@@ -0,0 +1,54 @@
...
+    "definitionType": "edmx"
+  },
+  "patches": [
+    {
+      "action": "merge",
+      "selector": {
+        "entityType": "A_BusinessPartnerType"
+      },
+      "data": {
+        "Core.Description": "Represents a natural or legal person involved in business transactions such as customers, vendors, or internal business units.",
```
swennemers: @Fannon , we should declare that we don't mess with the entity or property directly, but with annotations on the same. I haven't read it somewhere, but I like the approach...

Fannon: Yes, I think we need a statement that all metadata patches MUST be compatible to the original metadata spec. But I'm not sure if this is generally true, because maybe you want to use overlay to remove something you don't wan't to expose in the first place..

spec-extension/models/OrdOverlay.schema.yaml
```diff
@@ -31,6 +31,13 @@ description: |-
     - `jsonPath`: any JSON/YAML-based metadata file (including OpenAPI and MCP files).
     - `ordId`: ORD resource metadata level (patching ORD resources themselves).
 
+  Intended use and compatibility:
+    - Overlays MUST NOT introduce breaking / incompatible changes to the original API or metadata.
+    - Typical use-cases are non-breaking enrichments: adding extensions, annotations, or updating
+      non-critical properties such as descriptions, summaries, or display names.
+    - An exception is intentionally restricting exposure: an overlay MAY remove or hide elements
+      (e.g. operations, fields) that should not be visible in a given context or to a given audience.
```
Fannon: @swennemers  does this address your comments suggestions / concerns?

examples/overlay/edmx-business-partner.overlay.json
Fannon: I put this in now, see other comment where I tagged you

spec-extension/models/OrdOverlay.schema.yaml
swennemers: It is good to have these statement, but they don't address my issue above.

examples/overlay/edmx-business-partner.overlay.json
swennemers: @Fannon, your other comment doesn't address my issue here. There is a structural difference in OData EDMX between entityTypes and properties on one hand and annotations on the other hand. "Core.Description" is an annotation of the entityType in the target. Where is it documented, that the data section is targeting an annotation here? What if I would have a property that overlaps with an annotation.  
My concern is that we have here an EDMX specific merge logic/restriction to patching annotations and I don't see that specified.

spec-extension/models/OrdOverlay.schema.yaml
Fannon: What statement are you missing still?

swennemers: You have added it to the open topics section, it is the selector and annotations only question.

spec/v1/OrdOverlay.intro.md
```diff
@@ -0,0 +1,136 @@
...
+The full semantics of each action (`update`, `merge`, `remove`) are defined on the [`action`](#overlay-patch) field.
+
+Key points:
+- **`data` is required for `merge` and `update`**: `remove` omits `data` when the selected element should be removed entirely.
+- **`remove` semantics**:
+  - Omit `data` to remove the entire selected element.
+  - Provide `data` with `null`-valued properties to remove only those specific fields.
+- **`merge` behavior**: arrays are appended, not replaced. To fully replace an array, use two ordered patches — first `remove` the array field with `data: { "arrayField": null }`, then `merge` the new value.
+
+## Validation
```
Fannon, line 149: Make sure we have rules what is allowed to do in an overlay: 

- Must not change the target contract in incompatible ways ?

Typical allowed usage is to extend / enrich content 

In ORD target patch, we don't allow to change anything incompatible, overlays would be merged in by default.

spec-extension/models/OrdOverlay.intro.md
```diff
@@ -0,0 +1,136 @@
...
+### Via the ORD Configuration Endpoint
+
+Overlays can be listed directly in the [ORD Configuration Endpoint](../../spec-v1/index.md#ord-configuration-endpoint) under `openResourceDiscoveryV1.overlays`.
+This is the preferred approach for cross-cutting overlays that are not tied to a single resource, or when patching ORD resource metadata itself.
+
+```json
+{
+  "openResourceDiscoveryV1": {
+    "overlays": [
+      { "url": "/ord/overlays/my-api.overlay.json", "accessStrategies": [{ "type": "open" }] }
```
Fannon: Add `purpose` here, same way as we'd add them on definitions

https://github.com/open-resource-discovery/specification/pull/113

spec/v1/OrdOverlay.schema.yaml
```diff
@@ -0,0 +1,1054 @@
...
+    minLength: 1
+    description: |-
+      Optional description of the overlay document itself.
+
+      Notated in [CommonMark](https://spec.commonmark.org/) (Markdown).
+    examples:
+      - |
+        This overlay adds descriptive metadata for AI use-cases
+        without modifying the original resource definition.
+
```
Fannon, line 63: Add `purpose` for self-containment reasons? Tendency: no - one less thing to get inconsistent, to validate

spec-extension/models/OrdOverlay.schema.yaml
```diff
@@ -0,0 +1,1054 @@
...
+    type: object
+    title: Overlay Target
+    x-ums-type: "embedded"
+    description: |-
+      Optional context about the target being patched.
+      The target can refer to an ORD resource or to a referenced resource definition file.
+
+      When this object is present, at least one of `ordId`, `url`, `correlationIds`, or `definitionType`
+      MUST be provided so that consumers can identify what is being patched.
+      Exception: if all patches exclusively use [`selector.ordId`](#overlay-selector-by-ord-id) selectors, `target` may be omitted entirely.
```
vyshnavigadamsetti: What happens when there is no target specified ?  in this case would they have to use selector. ordid for reference ?  Is my understanding correct from the above statement that when a target is omitted all patches must use selector.ordid exclusively?

spec-extension/models/OrdOverlay.intro.md
```diff
@@ -0,0 +1,137 @@
...
+
+Overlays assume the target document is already valid for its native format.
+The merge tool does not fully re-validate target formats.
+After applying an overlay, validate the merged output with the corresponding format-specific tooling.
+
+## ORD Aggregator Expectations
+
+An ORD Aggregator MUST apply overlays that patch ORD resource metadata when building its ORD Discovery API and related indexes.
+This is necessary so that ORD-level overlay changes are reflected in discovery responses, filtering, searching, and similar aggregator behavior.
+
```
vyshnavigadamsetti, line 165: just  wanted to understand as there is no lastUpdate here in overlays, how would an aggregator detect changes?

spec-extension/models/OrdOverlay.schema.yaml
Fannon: We removed the selector.ordId for now. 

If there's no target specified, we assume that it is known via some other convention what get's merged together. That's not ideal, but we can't be completely opinionated that this happens. Maybe we should have a `sap:core:v1` policy rule that we always want an `target.ordId`

spec-extension/models/OrdOverlay.intro.md
Fannon: That's a good catch - you're right, we'd miss that. I think we have to add it

Fannon: Oh wait, we also don't have it on the config for ORD documents.. if we add it, we should add it for both - in a separate PR

spec/v1/OrdOverlay.intro.md
```diff
@@ -0,0 +1,165 @@
...
+  }
+}
+```
+
+### Attached to an ORD Resource
+
+Overlays can also be attached directly to an API or Event resource as a `resourceDefinitions` entry with type `ord:overlay:v1`.
+This keeps the overlay co-located with the resource it patches.
+
+A single resource can have multiple overlays with different `purpose` values.
```
Lars-H, line 36: What if there are two (or more) overlays with the same purpose on the same document? Is that a violation or does it mean that they are both applied leading to a single, new "overlayed" resource? If so, since patches are ordered sequences, there must be also an order of the overlays to get a total order of patches.

static/spec-v1/interfaces/ums/MetadataType/overlayselectorbyentityset.yaml
```diff
@@ -0,0 +1,26 @@
...
+  embedded: true
+  metadataProperties:
+    - name: 'id'
+      type: 'guid'
+      mandatory: true
+      unique: true
+      description: 'UMS ID (globally unique), used for relations.'
+    - name: 'entitySet'
+      type: 'string'
+      description: "Concept-level entity set identifier.\nSupported metadata formats:\n- `edmx` (OData v2/v4 CSDL XML): targets an EntitySet element inside EntityContainer.\n  May use the unqualified name (e.g. `Customers`) or namespace-prefixed name.\n- `csdl-json` (OData v4 CSDL JSON): targets a key with `$Collection: true` inside the\n  EntityContainer object in the namespace.\n\nUse `entitySet` when you need to patch EntityContainer-bound metadata such as\nCapabilities annotations (InsertRestrictions, UpdateRestrictions, etc.).\nFor patching the EntityType structure (Properties, NavigationProperties), use `entityType`."
```
Lars-H: > For patching the EntityType structure (Properties, NavigationProperties), use `entityType`.

I was a bit confused by this. Shouldn't this point to the "Overlay Selector By Property Type"?

Lars-H (approved):
@Fannon Just two smaller comments. Otherwise, it already looks good.

static/spec-v1/interfaces/ums/MetadataType/overlayselectorbyentityset.yaml
Fannon: Hm yes, I see why it's confusing. It's however also possible (and more convenient) to patch the properties of an entity type / object just by seleting this, and then providing the properties as deep merge. Some examples do this, it's also what SFSF did originally and it comes out a bit less verbose. 

Or not entirely sure this adressed the confusion already. For OData, I'm really not sure how mature this is as I'm not an OData expert myself and cannot judge it very well

pavelkornev:
@Fannon 

Please check rendering:

Screenshot 2026-05-12 at 17 10 43

spec/v1/OrdOverlay.intro.md
swennemers: we said no order guaranties, => no 2 overlays for the same purpose, if provided by different providers, aggregators can/must? reject this.

PR #92, opened 2026-03-19
Add releaseStatus: development

Fannon:
This came up in discussions with another team. They might want to already publish ORD resources that are still in development and wouldn't be for real use / delivery yet. But they might still need to already describe them, so the aggregator knows and other parts of the metadata ecosystem can learn about them.

### DIFF
--- a/spec/v1/Document.schema.yaml
+++ b/spec/v1/Document.schema.yaml
@@ -1038,6 +1038,11 @@ definitions:
 
           See [Lifecycle](../index.md#lifecycle) and [Compatibility](../concepts/compatibility.md) for more details.
         oneOf:
+          - const: development
+            description: |-
+              The resource is under active development and has not yet been released. The API contract is unstable and subject to
+              breaking changes at any time. Not intended for consumption outside of the development team.
+
           - const: beta
             description: |-
               The API contract has no stability guarantees. Breaking changes may occur at any time without notice or deprecation period.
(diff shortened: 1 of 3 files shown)


spec-extension/models/ORD_OVERLAY_CODE_REVIEW.md
```diff
@@ -0,0 +1,416 @@
+# ORD Overlay - Code Review & Beta Readiness Assessment
+
+**Date:** 2026-03-17
```
maiargu: I think his file does not belong to this PR and should be deleted

spec/v1/Document.schema.yaml
```diff
@@ -1038,6 +1038,11 @@ definitions:
 
           See [Lifecycle](../index.md#lifecycle) and [Compatibility](../concepts/compatibility.md) for more details.
         oneOf:
+          - const: development
```
maiargu, line 1052: what is the concrete difference between `development` and `beta`?
if there is a concrete difference, probably the description of both should be enhanced to clearly state it

Fannon: Will look to improve description.

No obligation to use this, it makes mostly sense if in development resources have to be published because a downstream consumer needs it again. Or the publishing app is the downstream consumer for this information again. 


Not a substitute for envirionment stages (dev, testing, qa, staging, prod).

Fannon: I slightly changed it, but if you have better ideas, we can improve it later again

vyshnavigadamsetti (approved):

PR #93, opened 2026-03-25
Allow integration dependencies to also refer to capabilitites

Fannon:
In case we use capabilitites to cover for agent skills, this would be needed. But I think it's generally sensible / useful. We already had an old TODO in the schema that this would make sense to add, too. 

### Added

- Added `CapabilityIntegrationAspect` definition to support capability dependencies in integration aspects
  - Capabilities can now be referenced as integration dependencies alongside API and Event Resources
  - Includes `ordId` and `minVersion` properties for capability integration aspects

### DIFF
--- a/examples/documents/document-integration-dependencies.jsonc
+++ b/examples/documents/document-integration-dependencies.jsonc
@@ -131,6 +131,17 @@
             },
           ],
         },
+        {
+          "title": "Field Extensibility Support",
+          "description": "Requires field extensibility capability for custom field integration",
+          "mandatory": true,
+          "capabilities": [
+            {
+              "ordId": "sap.s4:capability:fieldExtensibility:v1",
+              "minVersion": "2.0.0"
+            }
+          ]
+        },
       ],
     },
   ],
--- a/examples/documents/document-integration-dependencies.jsonc
+++ b/examples/documents/document-integration-dependencies.jsonc
@@ -131,17 +131,6 @@
             },
           ],
         },
-        {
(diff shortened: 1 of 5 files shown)


examples/documents/document-integration-dependencies.jsonc
```diff
@@ -131,6 +131,17 @@
             },
           ],
         },
+        {
```
Fannon: @swennemers Not sure this is a good example, we can also drop example until we have better one, e.g. via Agent Skill

src/generated/spec/v1/types/Document.ts
```diff
@@ -3243,6 +3247,22 @@ export interface EventResourceIntegrationAspectSubset {
    */
   eventType: string;
 }
+/**
+ * Capability related integration aspect
+ */
+export interface CapabilityIntegrationAspect {
```
maiargu, line 3253: I am not yet sure if this is indeed needed.

See existing `APIResourceIntegration` which can have a `subset` property which is of type `APIResourceIntegrationAspectSubset[]`.

And the definition of `APIResourceIntegrationAspectSubset` is:

Image

MCP capabilities are: tools, resources and prompts and all have a mandatory `name` property.

Fannon: Not sure I got this, this change is about allowing an integration dependency to not just point to an API (like MCP example you gave) but also a capability. For capabilitites, we don't have subsets atm.

vyshnavigadamsetti (approved):

PR #94, opened 2026-03-26
Add `files` property for Packages

pavelkornev:
Added a `files` property to Packages to allow linking of arbitrary files associated with packages, such as additional documentation in PDF format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ## [unreleased]
 
+- Added a `files` property to Packages to allow linking of arbitrary files associated with packages, such as additional documentation in PDF format.
+
 ## [1.14.1]
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ## [unreleased]
 
+### Added
+
+- Added `x-deprecated-in-version` and `x-deprecation-text` information to the exported JSON Schema
 - Added a `files` property to Packages to allow linking of arbitrary files associated with packages, such as additional documentation in PDF format.
 
 ## [1.14.1]
--- a/spec/v1/Document.schema.yaml
+++ b/spec/v1/Document.schema.yaml
@@ -449,6 +449,13 @@ definitions:
         items:
           $ref: "#/definitions/Link"
 
+      files:
+        type: array
+        description: |-
+          Generic list of files with arbitrary meaning and content. Meant to be used for linking PDFs, Word or similar content. This option MUST NOT be used for linking the actual metadata files like OpenAPI, AsyncAPI, CSN, etc.
+        items:
+          $ref: "#/definitions/File"
+
       licenseType:
         type: string
         minLength: 1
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ## [unreleased]
```
Fannon, line 12:
```suggestion
## [unreleased]

## [1.14.2]
```

If you already bump the version in package.json, you should add it here. Or we collect multiple PRs to the v1.14.2 release. Usually I create a separate `/release/v1.14.2` branch after merging the PRs together that go into it.

spec/v1/Document.schema.yaml
```diff
@@ -3709,6 +3716,63 @@ definitions:
 
   ############################################################################################################
 
+  File:
+    type: object
+    title: File
+    description: |
+      File that can be attached to packages or Open Resource Discovery resources.
```
```suggestion
      File that can be attached on ORD package level.
```

For now only on package level

CHANGELOG.md
pavelkornev: rolled back the version bump

spec/v1/Document.schema.yaml
```diff
@@ -3709,6 +3716,63 @@ definitions:
...
+          [URL](https://tools.ietf.org/html/rfc3986) of the link.
+
+          The file target MAY be relative or absolute.
+          If a relative URL is given, it is relative to the [`describedSystemInstance.baseUrl`](#system-instance_baseurl).
+          If an absolute URL is given, then it MUST be openly accessible.
+        examples:
+          - "/ord/v1/Ariba/Files/DeveloperPortal/SAP-Ariba-developer-portal.pdf"
+
+      description:
+        <<: *description
```
maiargu:
```suggestion
```
this line is not needed and can be deleted?

spec/v1/Document.schema.yaml
```diff
@@ -3709,6 +3716,63 @@ definitions:
...
+        description: |-
+          Full description, notated in [CommonMark](https://spec.commonmark.org/) (Markdown)
+        examples:
+          - |
+            Includes administrator, quick start, and authentication guides.
+
+      mediaType:
+        <<: *mediaType
```
maiargu, line 3780:
```suggestion
```
this line is not needed and can be deleted?

spec/v1/Document.schema.yaml
```diff
@@ -3709,6 +3716,63 @@ definitions:
 
   ############################################################################################################
 
+  File:
```
maiargu, line 3742: I just realised there exists already a `Link` entity and a `links` property in the spec, maybe an alternative to be considered?

Fannon: Not sure we should mix it up, Files have a special behavior where you don't just use the url as provided, you actually have to download and serve the file then under a new URL with the given media type. 

I think it's good to not mix both up into one thing.

spec/v1/Document.schema.yaml
Fannon: Good point, we inherit some basics (type: string, minLength: 1), but we also inherit an example that doesn't really fit then. Maybe we should generally not inherit descriptions.. I'll do separate PR to fix this

spec/v1/Document.schema.yaml
Fannon: It's needed, we want to inherit things like regexp. I would keep this one

spec/v1/Document.schema.yaml
Fannon: Ok, changed it to not inherit whole object, based on #98

vyshnavigadamsetti (approved):

Fannon (approved):

PR #98, opened 2026-03-27
Do not inherit descriptions via YAML Anchor tag

Fannon:
I think we accidentally inherited some examples / descriptions that were not really applicable. Let's not use YAML anchor inheritance for something as simple as `description`.

### DIFF
--- a/spec/v1/Document.schema.yaml
+++ b/spec/v1/Document.schema.yaml
@@ -385,7 +385,7 @@ definitions:
         examples:
           - "SAP S/4HANA Cloud, our next generation cloud ERP suite designed for in-memory computing."
 
-      description: &description
+      description:
         type: string
         minLength: 1
         description: |-
@@ -709,7 +709,8 @@ definitions:
           - "SAP LoB System 3 Cloud, cloud ERP suite designed for in-memory computing"
 
       description:
-        <<: *description
+        type: string
+        minLength: 1
         examples:
           - |
             This Consumption Bundle includes all SAP LoB System 3 Cloud APIs and Events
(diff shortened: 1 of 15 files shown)


spec/v1/Document.schema.yaml
```diff
@@ -385,7 +385,7 @@ definitions:
         examples:
           - "SAP S/4HANA Cloud, our next generation cloud ERP suite designed for in-memory computing."
 
-      description: &description
+      description:
         type: string
         minLength: 1
         description: |-
```
maiargu: probably this part is what would be nice to be reused, now it's missing, see missing comments in ts types

Fannon: Ok, I reused it now.

spec/v1/Document.schema.yaml
```diff
@@ -3692,9 +3714,10 @@ definitions:
           - "https://blogs.sap.com/2018/04/11/testing-of-s4hana-inventory-management-odata-apis-via-sap-api-hub/"
 
       description:
-        <<: *description
+        type: string
+        minLength: 1
         description: |-
```
maiargu, line 3719: here by intention a different text where this part 
```
      The description SHOULD not be excessive in length and is not meant to provide full documentation.
          Detailed documentation SHOULD be attached as (typed) links.
```
is missing?

maiargu: probably yes because this is the `link` property itself, ok fine for me

maiargu (approved):

spec/v1/Document.schema.yaml
Fannon: yes, here it's for Links, not sure it needs the full description or refer back to itself like this

PR #104, opened 2026-03-30
Add Agent.definitions

Fannon:
To be discussed if we need it, e.g. for Agent Extension Points Definition (which is independent from a particular exposed API of the agent). 

### Added

- Added `definitions` property to `Agent` resources to allow linking machine-readable agent definitions
  - Includes new `AgentDefinition` object with `type`, `mediaType`, `url`, `accessStrategies`, and `visibility` properties
  - The `type` field accepts any valid [Specification ID](./docs/spec-v1/index.md#specification-id) as an extensible enum
  - Similar to how API Resources, Event Resources, and Capabilities support definitions


Fannon:
Unsure if we should do it, honestly. It feels like we make it too easy to put the definition to the wrong place, as usually the detailed definition files are protocol dependent and therefore belong to the API. On agent level, we might get away from adding enough labels, e.g. what we started with `sap:ai:v1` now

Fannon:
There is currently no immediate requirement for Agent.definitions from Joule or UMS. Keeping this PR in draft until a concrete requirement emerges.

PR #105, opened 2026-03-30
Add Entity Type Definitions

Fannon:
### Added

- Added `definitions` property to `EntityType` resources to allow linking machine-readable entity type definitions
  - Includes new `EntityTypeDefinition` object with `type`, `mediaType`, `url`, `accessStrategies`, and `visibility` properties
  - The `type` field accepts any valid [Specification ID](./docs/spec-v1/index.md#specification-id)
  - **Important**: Entity Type definitions SHOULD usually be marked as `private` since they represent conceptual domain models that cannot be accessed directly. Consumers should interact with the data through related API Resources instead. Use the API Resource's `relatedEntityTypes` property to discover which APIs expose a particular entity type.


spec/v1/Document.schema.yaml
```diff
@@ -2056,6 +2056,24 @@ definitions:
         examples:
           - [{ "ordId": "sap.odm:entityType:WorkforcePerson:v1" }]
 
+      definitions:
+        type: array
+        x-introduced-in-version: "1.15.0"
+        x-feature-status: beta
+        description: |-
+          List of available machine-readable definitions, which describe the entity type's conceptual domain model in detail.
```
```suggestion
          List of available machine-readable definitions, which describe the entity type's internal model definitions in detail.
```

spec/v1/Document.schema.yaml
```diff
@@ -2056,6 +2056,24 @@ definitions:
...
+          See also [Resource Definitions](../index.md#resource-definitions) for more context.
+
+          Each definition is an alternative description format for the same entity type.
+          The same definition type MUST NOT be provided more than once, except with different `visibility`.
+
+          It is RECOMMENDED to provide definitions to enable machine-readable use cases.
+          If definitions are added or changed, the `version` MUST be incremented.
+
+          **Privacy Note**: Entity Type definitions SHOULD usually be marked as `private` visibility since they represent conceptual models that cannot be accessed directly. See `EntityTypeDefinition` for detailed guidance.
```
```suggestion
          **Privacy Note**: Entity Type definitions SHOULD usually be marked as `private` visibility since they are an internal implementation detail that cannot be accessed directly. See `EntityTypeDefinition` for detailed guidance.
```

spec/v1/Document.schema.yaml
```diff
@@ -2106,6 +2124,124 @@ definitions:
...
+        type: string
+        description: |-
+          Who is allowed to access the entity type definition. Defaults to `private`.
+
+          Entity Type definitions typically contain internal modeling details and SHOULD remain `private`.
+          Consumers interact with the data through related API Resources, not the conceptual model directly.
+
+          The visibility MUST be equal to or more restrictive than the Entity Type resource's visibility.
+
+          Only use `public` visibility when the conceptual model itself needs open access for standardization or documentation purposes.
```
```suggestion
          Entity Type definitions typically contain internal modeling details and SHOULD remain `private`.
          Consumers interact with the data through related API Resources, not the conceptual model directly.

          The visibility MUST be equal to or more restrictive than the Entity Type resource's visibility.

          Only use `public` visibility when the internal model definition itself needs open access for standardization or documentation purposes.
```

docs/spec-extensions/policy-levels/sap-core-v1.md
```diff
@@ -178,6 +178,11 @@ The following constraints apply in addition to the constraints defined in the [O
       - Consequently, the producing application cannot describe how they are eventually consumed.
 - SAP Event Catalogs SHOULD be validated via the SAP API Metadata Validator, using `sap:core:v1` compliance level.
 
+### Entity Type
+
+- For [Entity Types](../../spec-v1/interfaces/Document.md#entity-type):
+  - Entity Type definitions MUST have `visibility` set to either `private` or `internal`.
```
swennemers, line 184: @Fannon why?

spec/v1/Document.schema.yaml
```diff
@@ -2107,6 +2115,119 @@ definitions:
...
+        oneOf:
+          - const: public
+            description: |-
+              Publicly accessible.
+          - const: internal
+            description: |-
+              Can be accessed by applications and services from the same company / legal entity.
+          - const: private
+            description: |-
+              Can only be accessed by applications and services from the same installation / cluster.
```
```suggestion
              Can only be accessed by applications and services from the same internal organisation / installation / cluster, identified through the system namespace.
```

swennemers (approved):

PR #96, opened 2026-03-26
Add missing relations with custom relationType to Capability

Fannon:
### Added

- Added relationship properties to `Capability` for expressing relations to other ORD resources:
  - `relatedApiResources`: Array of `RelatedApiResource` objects to link capabilities to API resources
  - `relatedEventResources`: Array of `RelatedEventResource` objects to link capabilities to event resources
  - `relatedCapabilities`: Array of `RelatedCapability` objects to link capabilities to other capabilities
  - Each relationship object includes the ORD ID of the related resource and an extensible `relationType`
- Existing `relatedEntityTypes` property on `EntityType` has been extended to include `relationType` as well, for consistency and extensibility.

PR #110, opened 2026-04-02
Add missing relations with custom relationType to APIResource and EventResource

maiargu:
- Added relationship properties to ApiResource for expressing relations to other ORD resources:
  - relatedApiResources: Array of RelatedApiResource objects to link API resources to API resources
  - relatedEventResources: Array of RelatedEventResource objects to link API resources to event resources
- Added relationship properties to EventResource for expressing relations to other ORD resources:
  - relatedEventResources: Array of RelatedEventResource objects to link event resources to event resources
  - relatedApiResources: Array of RelatedApiResource objects to link event resources to API resources

related PR: https://github.com/open-resource-discovery/specification/pull/96

### DIFF
--- a/spec/v1/Document.schema.yaml
+++ b/spec/v1/Document.schema.yaml
@@ -1117,6 +1117,18 @@ definitions:
           - "1.2.3"
           - "2024.8.0"
 
+      relatedApiResources: &relatedApiResources
+        type: array
+        x-introduced-in-version: "1.14.3"
+        description: |-
+          Optional list of related API Resources.
+
+          Use this to indicate which APIs implement, expose, or are otherwise related to this capability.
+        items:
+          $ref: "#/definitions/RelatedApiResource"
+        examples:
+          - [{ "ordId": "sap.s4:apiResource:API_BUSINESS_PARTNER:v1", "relationType": "foo.bar:relationName" }]
+
       deprecationDate: &deprecationDate
         type: string
         format: date-time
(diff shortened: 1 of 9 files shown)


spec/v1/Document.schema.yaml
```diff
@@ -3122,28 +3146,12 @@ definitions:
...
-
-          Use this to indicate which APIs implement, expose, or are otherwise related to this capability.
-        items:
-          $ref: "#/definitions/RelatedApiResource"
-        examples:
-          - [{ "ordId": "sap.s4:apiResource:API_BUSINESS_PARTNER:v1", "relationType": "foo.bar:relationName" }]
 
       relatedEventResources:
-        type: array
+        <<: *relatedEventResources
```
Fannon, line 3161: @maiargu please check that the descriptions of the relation itself is general enough to be inheritited, e.g. not mention capability

src/generated/spec/v1/types/Document.ts
```diff
@@ -1605,6 +1635,24 @@ export interface EventResource {
    */
   systemInstanceAware?: boolean;
 }
+/**
+ * Defines a relation to an Event Resource (via its ORD ID).
+ */
+export interface RelatedEventResource {
```
Fannon: Why is this now created and not before?

spec/v1/Document.schema.yaml
```diff
@@ -1117,6 +1117,18 @@ definitions:
           - "1.2.3"
           - "2024.8.0"
 
+      relatedApiResources: &relatedApiResources
```
Fannon, line 1120: We also need `relatedEventResources` here, so we should always have both on both APis and Events

src/generated/spec/v1/types/Document.ts
maiargu: it was created also before, just the place of the interface definition switched some lines above in the file scope

spec/v1/Document.schema.yaml
maiargu: added

spec/v1/Document.schema.yaml
maiargu: exchanged the word `capability` with `entity` to have still a reusable description

Fannon (approved):

PR #112, opened 2026-04-07
Add Agent Governance ORD model extension (alpha)

vyshnavigadamsetti:
## Summary

Introduces `AgentGovernance` as an ORD model extension that provides governance metadata for AI agents and MCP servers.

This extension intentionally separates governance concerns from the core agent/MCP resource lifecycle — allowing AI Agent Hub to publish governance data independently from the team that registered the original resource in ORD.

**Phase 1 fields:**
- `mainContact` — name + email of the customer-side contact (EU AI Act compliance, incident response, cost allocation)
- `verified` — flag indicating whether an agent/MCP server is approved for production use

## Open Questions for Reviewers

1. **`verified` ownership** — Who should set this flag? Candidates are AI Agent Hub and Integration Suite.  
2. **`title` / `lastUpdate`** — Are these needed on AgentGovernance, or is `ordId` sufficient for tracking changes?
3. **`contactEmail` description** — Currently says "if only one attribute can be captured, it MUST be this one" but both `name` and `contactEmail` are now required. Is this still accurate?
4. **`mainContact` required constraint** — `mainContact` is marked as required, but there may be cases where a governance contact is not yet known at the time the record is first created. Should `mainContact` be optional to support this scenario?


spec-extension/models/AgentGovernance.schema.yaml
```diff
@@ -2,9 +2,165 @@ $schema: "https://json-schema.org/draft/2020-12/schema"
...
+      allowed to run in production environments.
+    examples:
+      - true
+      - false
+
+required:
+  - agentGovernance
+  - ordId
+  - target
+  - mainContact
```
Fannon: I don't think this can be mandatory, as it's not always given. If there are more properties related to Agent governance, I guess all of them will be optional ?

spec-extension/models/AgentGovernance.schema.yaml
```diff
@@ -2,9 +2,165 @@ $schema: "https://json-schema.org/draft/2020-12/schema"
...
+      When retrieved from an ORD aggregator, `lastUpdate` will be reliable there and reflect either the provider based update time or the aggregator processing time.
+      Therefore consumers MAY rely on it to detect changes to the metadata and the attached resource definition files.
+
+      If the resource has attached definitions, either the `version` or `lastUpdate` property MUST be defined and updated to let the ORD aggregator know that they need to be fetched again.
+
+      Together with `perspectives`, this property SHOULD be used to optimize the metadata crawling process of the ORD aggregators.
+    examples:
+      - "2022-12-19T15:47:04+00:00"
+
+  target:
```
Fannon, line 67: I think this can be simplified towards a simple `agent`, where the value is an ORD ID of the agent?

spec-extension/models/AgentGovernance.schema.yaml
vyshnavigadamsetti: in the agent ownership adr, we have a statement which says maincontact is required and within that statement it is also menitoned contact email is the critical field if only one can be captured, it must be contact email

spec-extension/models/AgentGovernance.schema.yaml
vyshnavigadamsetti: target can be either an agent or an apiResource (MCP server), so using agent as the field name could be misleading?

Fannon: I think only Agent should be the target, as govenance happens on conceptual Agent level, not on API level I think. 

This is also not MCP Server related. If we want to generalize it more, then I guess we shouldn't call it "AgentGovernance", but maybe just "Governance". In this case, we should also think about adding this to APIs, Events, Data Products (especially). @swennemers do you have an opinion if that makes sense?

Having Agent Governance specifically would allow us to add later more field which may only be applicable to Agents. But right now the intended attributes would all be more generic afaik.

spec-extension/models/AgentGovernance.schema.yaml
Fannon: That ADR was only about adding main contacts, so if you add them it's mandatory. But adding them overall or doing agent governance as such is an optional task. If we now combine this requirement with other requirements, they cannot all be mandatory at the same time

vyshnavigadamsetti: Just to clarify  did you mean 'name' within mainContact can be optional, but mainContact itself (with email) should still be required when creating a governance record?

HelgeCD:
One thing that I am wondering a bit is if we need something like a central config and if this should also be part of the model extension.

I think we need to define when this governance needs to apply. i.e. is the verification flag mandatory in an environment or is it optional. In an environment with strict rules an agent that has no verification flag (or AgentGovernance File) should never be visible, but in an environment with more relaxed rules this could be ok. This could be modeled as an additional file with for example a ```mandatoryAgentVerification``` flag.

Really not sure if this something that should be modeled on metadata level so just throwing the idea out there.

vyshnavigadamsetti:
My understanding is that `verified` is primarily for Agent Gateway filtering. Whether it's mandatory in an environment would be a runtime concern for Agent Gateway to own, not the schema. The schema defaults `verified` to `false`, so agents without a governance record are unverified by default.

spec-extension/models/AgentGovernance.schema.yaml
Fannon: Ok, I changed my mind, having this as an object is 1) more flexible for the future and 2) more in sync how it works in the Overlay extension, too. Let's keep it

spec-extension/models/AgentGovernance.schema.yaml
Fannon: There are probably more reasons to add governance information to an Agent, not just adding a mainContact. If you have another reason to add information, but you don't have / want to add an mainContact (or someone else adds it), having it as mandatory would block you.

Or we could decide that having mainContact is so essential, it should always be supplied when doing governance - then we can have it mandatory

spec-extension/models/AgentGovernance.schema.yaml
vyshnavigadamsetti: One thing I was  also thinking about is the naming. Agent Governance could be misleading if we have more things in scope.  should we consider something like Agentecosystem governance if AI Governance or Governance could be too broad ?

spec-extension/models/AgentGovernance.schema.yaml
vyshnavigadamsetti: My understanding is that the core governance use case is when a customer/EA sees an agent in AI Agent Hub, they need someone to reach out to. A governance record without a contact might not really serve that purpose.

spec-extension/models/AgentGovernance.schema.yaml
```diff
@@ -2,9 +2,165 @@ $schema: "https://json-schema.org/draft/2020-12/schema"
...
+  mainContact:
+    $ref: "#/definitions/MainContact"
+
+  verified:
+    type: boolean
+    default: false
+    description: |-
+      Indicates whether this agent or MCP server has been verified and approved for use.
+
+      This flag is set by a governance system (e.g. Integration Suite or AI Agent Hub). The exact
```
perbernhardt: This would only be AI Agent Hub. Integration Suite does not set the verification flag, it only interprets it (for MCP servers).

spec-extension/models/AgentGovernance.schema.yaml
perbernhardt: I am unsure about the "contact" topic. We should only have AI Agent Hub ever set one of these governance extension records. And I am not sure, whether we would be the ones to define the mainContact (name/email) in the first place. E.g. when a user builds a new agent in Joule Studio, then Joule Studio should write this user as a contact into UMS.
It might only make sense for AI Agent Hub to overwrite/change the responsibility later, but then this is probably only an optional thing.
So in short: If we keep this in the governance extension, then it should be optional IMHO.

spec-extension/models/AgentGovernance.schema.yaml
```diff
@@ -2,9 +2,165 @@ $schema: "https://json-schema.org/draft/2020-12/schema"
...
+        type: string
+        maxLength: 255
+        pattern: ^([a-z0-9]+(?:[.][a-z0-9]+)*):(agent|apiResource):([a-zA-Z0-9._\-]+):(v0|v[1-9][0-9]*)$
+        description: |-
+          [ORD ID](../../spec-v1/index.md#ord-id) of the governed agent or MCP server.
+        examples:
+          - "sap.foo.bar:agent:disputeAgent:v1"
+          - "sap.foo.bar:apiResource:mySapMcpServer:v1"
+    required:
+      - ordId
```
perbernhardt, line 126: Remembering the "Agent Identifier" CPA decision, I think "ordId" is not enough. IIRC, the ordId would be similar to an "agent type". But we need another ID (I think it said tenant ID), to differentiate agents of the same type for e.g. different customers.
If this is true, then we also need both IDs in this extension to be able to govern agents per customer, no?

spec-extension/models/AgentGovernance.schema.yaml
```diff
@@ -2,9 +2,165 @@ $schema: "https://json-schema.org/draft/2020-12/schema"
...
+    type: object
+    title: Main Contact
+    x-ums-type: "embedded"
+    description: |-
+      The main contact responsible for this agent or MCP server from a governance perspective.
+
+      This is a customer-side contact — the person an Enterprise Architect can reach out to
+      when reviewing agents in the SAP AI Agent Hub. It is NOT the SAP IP owner of the agent.
+
+      - For **SAP Standard agents**: set by the IT Admin during provisioning in SAP for Me.
```
perbernhardt, line 138: Ok, reading this, I think this extension is the wrong place for this information. The extension should typically only be written/owned by AI Agent Hub. As you write, in many situations, another system would be able to define the contact (via SAP for Me or person deploying the agent). Shouldn't this be part of the main agent record in UMS then?

spec-extension/models/AgentGovernance.schema.yaml
```diff
@@ -2,9 +2,165 @@ $schema: "https://json-schema.org/draft/2020-12/schema"
...
+        description: Full name of the main contact person.
+        examples:
+          - "Jane Doe"
+      contactEmail:
+        type: string
+        format: email
+        description: |-
+          Email address of the main contact person.
+
+          This is the most critical field — if only one attribute can be captured, it MUST be this one.
```
perbernhardt, line 159: Another argument to have this as part of the main agent record. AI Agent Hub governance is an optional activity, not all customers will do this for all agents. Hence, we should not expect this to be a source for a mandatory compliance field.

spec-extension/models/AgentGovernance.schema.yaml
perbernhardt: I can live with AgentGovernance too. We already have the same issue with the "AI Agent Hub": The name only refers to agents though we also support managing your MCP servers.

perbernhardt:
Left some comments. Does it make sense to jump on a call to get this merged?

spec-extension/models/AgentGovernance.schema.yaml
```diff
@@ -2,9 +2,165 @@ $schema: "https://json-schema.org/draft/2020-12/schema"
...
+    maxLength: 255
+    pattern: ^([a-z0-9]+(?:[.][a-z0-9]+)*):(agentGovernance):([a-zA-Z0-9._\-]+):(v0|v[1-9][0-9]*)$
+    description: |-
+      Unique [ORD ID](../../spec-v1/index.md#ord-id) for this Agent Governance record.
+
+      The resource type segment MUST be `agentGovernance`.
+    examples:
+      - "sap.foo:agentGovernance:disputeAgent:v1"
+
+  title:
```
swennemers: do we really need a title? Wouldn't UIs always merge the information with the agents title?

spec-extension/models/AgentGovernance.schema.yaml
swennemers: We might get a data product governance in the future as well... one should discuss with the DSP people, but it is currently no priority.
So would we be fine with an AgentGovernance and a DataProductGovernance eventually?

Why don't we just use the Overlay concept for adding these information?

spec-extension/models/AgentGovernance.schema.yaml
perbernhardt: Agreed

perbernhardt:
> One thing that I am wondering a bit is if we need something like a central config and if this should also be part of the model extension.
> 
> I think we need to define when this governance needs to apply. i.e. is the verification flag mandatory in an environment or is it optional. In an environment with strict rules an agent that has no verification flag (or AgentGovernance File) should never be visible, but in an environment with more relaxed rules this could be ok. This could be modeled as an additional file with for example a `mandatoryAgentVerification` flag.
> 
> Really not sure if this something that should be modeled on metadata level so just throwing the idea out there.

By now, I think that this should be part of the system that interprets the verification flag. As an example: Integration Suite will use the verification flag to suspend execution of unverified MCP servers. Whether a given Integration Suite instance (e.g. prod, dev, certain region etc.) actually considers the flag is part of the configuration of that very system. They also have additional complexity there, because they need to handle the situation that a lot of MCP servers are already defined before the verification is introduced. That is why they will even support a gradual migration (e.g. whitelisting MCP servers that have been defined earlier etc.). All of this is very Integration Suite specific and hence not part of this generic UMS extension.
Along the same lines, I would expect this to be configurable on certain agent runtimes etc.

perbernhardt:
> My understanding is that `verified` is primarily for Agent Gateway filtering. Whether it's mandatory in an environment would be a runtime concern for Agent Gateway to own, not the schema. The schema defaults `verified` to `false`, so agents without a governance record are unverified by default.

Agreed. As explained in more detail in my [previous comment](https://github.com/open-resource-discovery/specification/pull/112#issuecomment-4236687680), I see this as a concern/configuration on the runtime, too.

spec-extension/models/AgentGovernance.schema.yaml
vyshnavigadamsetti: Agreed, as the governance record might always be shown alongside the agent it governs, so a separate title is redundant. Will remove it.

spec-extension/models/AgentGovernance.schema.yaml
vyshnavigadamsetti: fixed in  '98190b5'

perbernhardt:
@vyshnavigadamsetti as discussed, let's also add the second boolean flag "miningActivated". Thank you!

spec-extension/models/AgentGovernance.schema.yaml
vyshnavigadamsetti: fixed in 'fb73c2e6'

vyshnavigadamsetti:
> @vyshnavigadamsetti as discussed, let's also add the second boolean flag "miningActivated". Thank you!

Made changes, please verify the description. '[a3bc5979](https://github.com/open-resource-discovery/specification/pull/112/changes/a3bc5979ac13ab64a00197856e8e41522d104a48)'

static/spec-v1/interfaces/AgentGovernance.schema.json
```diff
@@ -1,10 +1,134 @@
...
+      "description": "Indicates whether this agent or MCP server has been verified and approved for use.\n\nThis flag is set by the AI Agent Hub. Consumers such as Integration Suite and the\nAgent Gateway can use this flag to filter which agents are allowed to run in\nproduction environments.",
+      "examples": [
+        true,
+        false
+      ]
+    },
+    "miningActivated": {
+      "type": "boolean",
+      "default": false,
+      "description": "Indicates whether mining has been activated for this agent or MCP server.\n\nThis flag is set by the AI Agent Hub alongside `verified`. While `verified` indicates\napproval for use, `miningActivated` controls whether the agent's usage data is actively\ncollected and analyzed.",
```
perbernhardt, line 70: Mining is only relevant for agents, not for MCP servers.

perbernhardt: I would also change the second part:

While `verified` indicates\napproval for use, `miningActivated` controls whether the agent's telemetry is further analyzed.

spec-extension/models/AgentGovernance.schema.yaml
```diff
@@ -2,9 +2,162 @@ $schema: "https://json-schema.org/draft/2020-12/schema"
...
+    examples:
+      - "2022-12-19T15:47:04+00:00"
+
+  target:
+    $ref: "#/definitions/AgentGovernanceTarget"
+
+  mainContact:
+    $ref: "#/definitions/MainContact"
+
+  verified:
```
Fannon, line 73: I would still recommend to rename this to "approved" to reflect better what this is used for. This does not imply that you have to call it like this on UI, if LeanIX want's they can call it verified there. 

> Verified means checked for accuracy, truth, or compliance through evidence (technical confirmation), while Approved means formally authorized or accepted (administrative permission). Verified ensures correctness, whereas Approved signals permission to proceed

So if I understand the intended usage of it, it's actually for approval, not verification. Those are two differen things, we should not mix up. We might also get a verified flag at some point, which could be set .e.g. by code scanning / technical verification tools.

perbernhardt: I disagree. We discussed for a long time & with customers how to call the functionality/flag in the product. In terms of "ubiquitous language" really dislike calling it differently. This will only confuse people. My understanding is, that this extension is not an "official / public" schema but really the means to define something more proprietary next to it. I also understand that we (AI Agent Hub / LeanIX) would be owning this. So why wouldn't we go with our naming?

Fannon: If we make this a public ORD extension, it should not have proprietary / special meaning but more generalized. I would then strongly propose to not go against established semantics / meanings of existing IT terminology.

In this case you can't own it alone, you'd have to work and agree with a generalized understanding that also makes sense outside of a single product context. But I'm now leaning towards not standardizing it like this, it may be easier to make this a product-specific extension.

Fannon:
We decided to not standardize this at the moment. We'll remove the AgentGovernance model extension for now.

PR #113, opened 2026-04-13
Add definition.purpose as Concept ID

Fannon:
### Added

- Added `purpose` property to resource definitions (`ApiResourceDefinition`, `EventResourceDefinition`, `CapabilityDefinition`)
  - Describes the intended purpose or role of the definition (e.g., `ord:ai-enrichment` for AI-optimized definitions)
  - Allows multiple definitions of the same `type` when they serve different purposes
  - The combination of `type`, `purpose`, and `visibility` MUST be unique within a resource's definitions list
- Documented `ord` as a reserved vendor namespace for ORD specification-defined values in extensible enums


swennemers (approved):

spec/v1/Document.schema.yaml
```diff
@@ -2633,6 +2635,32 @@ definitions:
         examples:
           - "sap:custom-definition-format:v1"
 
+      purpose: &resourceDefinitionPurpose
```
Fannon, line 2648: TODO: Add `purpose` also to ORD Overlay, so it's self contained and there if provided via ORD config directly

Fannon: TODO: Make sure we explain ORD Aggregator merging overlay - the resulting definitions should inherit the purpose and visibility of the overlay

PR #116, opened 2026-04-14
Recommend labels key to be Concept ID to clarify ownership and avoid conflicts

Fannon:
### Changed

- Added recommendation to use [Concept IDs](https://open-resource-discovery.org/spec-v1#concept-id) as `labels` keys to indicate ownership and avoid naming conflicts.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -2,8 +2,5 @@
   "contextFiles": [
     "AGENTS.md",
     "package.json"
-  ],
-  "hooks": {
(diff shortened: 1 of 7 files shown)


.claude/settings.json
HelgeCD, line 1: looks fine overall, just to be sure: was this change to the claude settings an intentional change?

Fannon: yes, this was wrongly committed in, and I remved it.

PR #118, opened 2026-04-24
Clarify system-perspective fallback from system-version

Fannon:
- Clarification: Consolidated and clarified the [static perspective resolution](https://open-resource-discovery.org/spec-v1/concepts/perspectives#static-perspective-resolution) algorithm for aggregators. When no specific version is requested, explicit `system-type` perspective data takes precedence; if unavailable, the aggregator SHOULD derive it from the latest `system-version`. Previously this behavior was scattered and only stated as MAY.

### DIFF
--- a/docs/spec-v1/concepts/perspectives.md
+++ b/docs/spec-v1/concepts/perspectives.md
@@ -137,7 +137,7 @@ If the aggregator supports both static and dynamic perspectives:
 
 - The ORD aggregator MUST be able to aggregate and store all perspectives (`system-type`, `system-version`, and `system-instance`) at the same time.
 - In its ORD Discovery API for consumers, it needs to implement the inheritance / fallback behavior:
-  - When `system-instance` is requested but not available, fall back to `system-version`.
+  - When `system-instance` is requested but not available, fall back to `system-version`, when this is not available fall back to `system-type`.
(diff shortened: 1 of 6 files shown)


docs/spec-v1/concepts/perspectives.md
```diff
@@ -116,27 +127,30 @@ If the system has dynamic metadata, it should describe both perspectives complet
...
 
 If the aggregator supports both static and dynamic perspectives:
 
-- The ORD aggregator must be able to aggregate and store all perspectives (`system-type`, `system-version`, and `system-instance`) at the same time.
+- The ORD aggregator MUST be able to aggregate and store all perspectives (`system-type`, `system-version`, and `system-instance`) at the same time.
 - In its ORD Discovery API for consumers, it needs to implement the inheritance / fallback behavior:
-  - When `system-instance` is requested but not available, fall back to `system-version`
-  - When `system-version` is requested but not available (or version is unknown), fall back to `system-type`
-  - The aggregator MAY derive `system-type` from the latest `system-version` if not explicitly provided
+  - When `system-instance` is requested but not available, fall back to `system-version`.
```
Fannon:
```suggestion
  - When `system-instance` is requested but not available, fall back to `system-version`, when this is not available fall back to `system-type`.
```

makpac (approved):

PR #119, opened 2026-04-27
Add relationType "same-as" for related* relationTypes

Fannon:
### Added

- Added standardized `same-as` relation type for `relatedEntityTypes`, `relatedApiResources`, `relatedEventResources`, and `relatedCapabilities`, indicating that the source and target resource are the same or equivalent.


swennemers (commented):
I doubt we have a use case for this relationship type so far. 
The use cases I am aware of, should either be successors or should be modelled using a shared namespace => same ordId

PR #120, opened 2026-04-27
Fix labels key regexp to actually allow the characers as recommended Concept ID

Fannon:
### Fixed

- Fixed `labels` key validation pattern to also allow `:` and `/` characters, which are needed when using [Concept IDs](https://open-resource-discovery.org/spec-v1#concept-id) as label keys (as recommended).


Fannon:
@maiargu can you check why spec-toolkit didn't notice this issue? The example values are usually validated against the JSON Schema it is an example of - I would have thought this kind of problem we'd detect

PR #121, opened 2026-04-28
Docs/implementing ord natively

Fannon:
Preview: https://open-resource-discovery.github.io/pr-preview/specification/pr-121/spec-v1/concepts/implementing-ord-natively

### Added

- Added [Implementing ORD Natively](https://open-resource-discovery.org/spec-v1/concepts/implementing-ord-natively) guide

### DIFF
--- a/docs/spec-v1/concepts/implementing-ord-natively.md
+++ b/docs/spec-v1/concepts/implementing-ord-natively.md
@@ -45,8 +45,8 @@ const ordConfig = {
   },
 };
 
-const ordDocument = require("../../documents/document-1.json");
-const openApiDocument = require("../nginx-no-auth/metadata/astronomy-v1.oas3.json");
+const ordDocument = require("./ord-document.json"); // your ORD document
+const openApiDocument = require("./astronomy-v1.oas3.json"); // a referenced resource definition
 
 http
   .createServer((req, res) => {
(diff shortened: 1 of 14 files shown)


docs/spec-v1/concepts/implementing-ord-natively.md
```diff
@@ -0,0 +1,342 @@
...
+If a resource does not exist for a tenant, leave it out of that tenant's ORD document.
+Use `disabled` only when the resource exists for that tenant but is currently not available for consumption.
+
+For tenant-aware ORD, make sure the implementation clearly does these things:
+
+- advertise static and dynamic perspectives separately
+- resolve the tenant before generating `system-instance` metadata
+- map aggregator tenant IDs to local tenant IDs deliberately
+- generate a complete tenant-specific ORD document, not a delta
+- generate tenant-specific resource definitions with the same tenant context
```
vyshnavigadamsetti, line 303: what would happen when a tenant's model changes between crawls do they need to update `version` or `lastUpdate` on the affected resource so aggregators will refetch the resource definition? does this case even exists ?

swennemers: @vyshnavigadamsetti , this use case exists. And you have to update the lastUpdate. You MUST not change the version for e.g. custom fields as they version is bound to the resource lifecycle (SAP deliveries) and the custom field is bound to customer changing things independently. => lastUpdate to change.

vyshnavigadamsetti: then can we add something like 
- when a tenant's model changes between crawls, update lastUpdate on the affected resource so aggregators can re-fetch the resource definition.

Fannon: I added a checklist item for it, good point!

PR #125, opened 2026-05-01
Add root level baseUrl for provider system and clarifications around aggregator handling of relative URLs

Fannon:
### Added

- Added optional root-level `baseUrl` property to ORD Documents (analogous to the existing root-level `baseUrl` in the ORD Configuration).
  This property represents the base URL of the **ORD provider** and is used to resolve relative URLs to metadata files within the document (e.g., `resourceDefinitions[].url`).
  Particularly important in push, offline, and self-contained document scenarios, and REQUIRED when the provider and the described system differ.

### Changed

- Added a formal [Relative URL Resolution](https://open-resource-discovery.org/spec-v1#relative-url-resolution) section to the spec index, documenting:
  - Three URL reference types: absolute, root-relative (leading `/`), and document-relative (`./`, `../`, bare path)
  - Root-relative resolution follows RFC 3986 and resolves against the base URL *including its path component*
- Clarified the two-level base URL model:
  - Root-relative **metadata file** URLs (`resourceDefinitions[].url`, `overlayDefinitions[].url`, `File.url`, resource/data product links) resolve against the **provider base URL**
  - Root-relative **entry point** URLs resolve against `describedSystemInstance.baseUrl` (described system base URL)
  - In the common case where provider and described system are the same, both values are identical and the distinction has no practical effect
- Defined the resolution fallback order for the provider base URL (applied by ORD aggregators):
  1. Document root `baseUrl` (if present) — takes precedence
  2. In pull scenarios: the URL the ORD document was fetched from
  3. `describedSystemInstance.baseUrl` — backward-compatibility fallback for documents predating this version
- Clarified the resolution order for the described system base URL:
  1. Aggregator-authoritative URL — aggregators MAY prefer their own landscape knowledge (e.g., from service discovery) over the document-provided value
  2. `describedSystemInstance.baseUrl` — MUST be provided when the base URL is not otherwise known to the aggregator (push, offline, self-contained scenarios)
- Updated all relative URL descriptions across schema, TypeScript types, and UMS interface files to reference the document root `baseUrl` instead of the (previously incorrect) `describedSystemInstance.baseUrl`


swennemers (approved):

mlakov (approved):
LGTM

docs/spec-v1/index.md
```diff
@@ -310,6 +310,78 @@ Additional information or categorization can be added through the generic `Label
...
+Relative URL resolution follows [RFC 3986 (Uniform Resource Identifier: Generic Syntax)](https://datatracker.ietf.org/doc/html/rfc3986#section-5), with the base URI determined as described below.
+
+##### URL Reference Types
+
+ORD recognizes three types of URL references:
+
+| Pattern | Type | Resolved against |
+|---|---|---|
+| `https://example.com/path` | Absolute URL | Used as-is — no resolution needed |
+| `/path/file.json` | Root-relative (leading slash) | The applicable **base URL** (see below) |
```
pavelkornev, line 327: What is `baseUrl` is not defined? Where the "root-relative" should be resolved against? The domain name where the ORD document is served from?

In case of some ABAP system or some non-standard ORD implementation the ORD metadata might be in some deep folder structure, e.g.
`https://xxx-732-api.example.com/sap/bc/http/sap/ord_configuration/`

Would it mean that without `baseUrl` the resolution of
`/foo/bar/openapi.json`
should go to
`https://xxx-732-api.example.com/foo/bar/openapi.json`

?

docs/spec-v1/index.md
```diff
@@ -310,6 +310,78 @@ Additional information or categorization can be added through the generic `Label
...
+
+> **Note:** Per RFC 3986, `path/file.json` (no leading slash, no dot prefix) is equivalent to `./path/file.json` — both are document-relative.
+
+##### Base URL for Root-Relative URLs
+
+The base URL used for resolving root-relative URLs (leading slash) depends on **what** the URL references, because two different systems may be involved:
+
+| URL type | Resolved against | Declared via |
+|---|---|---|
+| **Metadata files**: `resourceDefinitions[].url`, `links[].url`, and similar file references | Provider base URL | Document root `baseUrl` |
```
pavelkornev: Are we sure `links[].url` can be relative?

pavelkornev:
>describedSystemInstance.baseUrl — backward-compatibility fallback for documents predating version 1.15 that do not provide a root baseUrl.

What does it mean for documents below 1.15? I would expect we have this backward compatibility check in 1.x unconditionally. If `baseUrl` is not specified in the 1.x documents (even above 1.15), this fallback should apply.

CHANGELOG.md
```diff
@@ -16,6 +16,25 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 - Added `correlationIds` to package
 - Added [Implementing ORD Natively](https://open-resource-discovery.org/spec-v1/concepts/implementing-ord-natively) guide
+- Added optional root-level `baseUrl` property to ORD Documents (analogous to the existing root-level `baseUrl` in the ORD Configuration).
+  This property represents the base URL of the **ORD provider** and is used to resolve relative URLs to metadata files within the document (e.g., `resourceDefinitions[].url`, `links[].url`).
```
Fannon: Not sure about links, but files should be

CHANGELOG.md
```diff
@@ -16,6 +16,25 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
...
+
+### Changed
+
+- Specified that relative URL resolution follows [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-5) semantics and documented the three URL reference types:
+  - **Absolute URLs** (`https://...`) — used as-is.
+  - **Root-relative URLs** (leading slash, e.g., `/path/file.json`) — resolved against the applicable base URL including its path component.
+  - **Document-relative URLs** (`./`, `../`, or bare path without leading slash) — resolved relative to the current document's URL per RFC 3986. This is particularly useful for static ORD providers serving files from a file system or git repository.
```
Fannon: Consider dropping this feature to simplify ORD Aggregators, so they only have to support first two options.

docs/spec-v1/index.md
Fannon: The aggregator might already know the baseUrl of a tenant / system, it might actually know it better than the system itself in case there are proxies / gateways used. 

Yes, it would then go to `https://xxx-732-api.example.com/foo/bar/openapi.json`

CHANGELOG.md
Fannon: Decided to keep it, as we already allowed it before and it would be a breaking change to suddenly stop allowing it

docs/spec-v1/index.md
Fannon: So far we allowed it but it was a RECOMMENDATION to make them absolute. 

But I had the same question. If they are relative.. what are they relative to? Probably baseUrl / describedSystemInstance.baseUrl in that order. But I would expect that most links would point to different locations and they have to be absolute anyways

PR #126, opened 2026-05-04
Add Package `correlationIds`

maiargu:


spec/v1/Document.schema.yaml
```diff
@@ -361,6 +361,23 @@ definitions:
...
+
+          If a "part of" relationship needs to be expressed, use the `partOfGroups` assignment instead.
+
+          MUST be a valid [Correlation ID](../index.md#correlation-id).
+        items: &correlationId
+          type: string
+          pattern: ^([a-z0-9]+(?:[.][a-z0-9]+)*):([a-zA-Z0-9._\-\/]+):([a-zA-Z0-9._\-\/]+)$
+          maxLength: 255
+        examples:
+          - ["sap.s4:communicationScenario:SAP_COM_0008"]
```
pavelkornev:
```suggestion
          - ["sap.s4:communicationScenario:SAP_COM_0008", "sap.businesshub:package:SAPS4HANACloudBusinessEvents"]
```

pavelkornev (approved):

spec/v1/Document.schema.yaml
Fannon:
```suggestion
          - ["sap.businesshub:package:SAPS4HANACloudBusinessEvents"]
```

Only this one makes sense

Fannon (approved):

spec/v1/Document.schema.yaml
maiargu: correlationId `conceptName` can be any string https://open-resource-discovery.org/spec-v1#correlation-id

Fannon: Yes, but the examples should correlate it to a concept that is related / mappable. A package does not relate in any way to a communication scenario. If we want to share an example, we could go for a completely generic example like `foo:bar:concept:targetId`?

maiargu: used generic example `"foo.bar:conceptName:localIdentifier"`

PR #129, opened 2026-05-09
Make ORD ID major version fragment sync only RECOMMENDED. Create detail page.

Fannon:
### Changed

- Relaxed the rule that `<majorVersion>` in the ORD ID and the major of `version` MUST be identical to a SHOULD (recommended). Strict enforcement created an unresolvable conflict with the ORD ID stability requirement: when a provider bumps the semver major without creating a new resource — whether because no breaking change occurred, or because they simply didn't follow best practice — the old MUST rule would have required changing the published ORD ID of an existing resource, breaking every downstream consumer that references it. Even when a breaking change did happen, forcing an ID change compounds the provider's mistake rather than containing it. The `<majorVersion>` is now explicitly defined as tracking *breaking API changes*, with `version` major as a strong signal that should be kept in sync but does not override ID stability. Validators SHOULD warn on a mismatch but MUST NOT treat it as a hard failure.
- Moved the Versioning and Lifecycle content into a dedicated [Versioning and Lifecycle](https://open-resource-discovery.org/spec-v1/concepts/versioning-and-lifecycle) concept page with expanded guidance.


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ## [unreleased]
 
+### Changed
+
+- Relaxed the rule that `<majorVersion>` in the ORD ID and the major of `version` MUST be identical to a SHOULD (recommended). Strict enforcement created an unresolvable conflict with the ORD ID stability requirement: when a provider bumps the semver major without creating a new resource — whether because no breaking change occurred, or because they simply didn't follow best practice — the old MUST rule would have required changing the published ORD ID of an existing resource, breaking every downstream consumer that references it. Even when a breaking change did happen, forcing an ID change compounds the provider's mistake rather than containing it. The `<majorVersion>` is now explicitly defined as tracking *breaking API changes*, with `version` major as a strong signal that should be kept in sync but does not override ID stability. Validators SHOULD warn on a mismatch but MUST NOT treat it as a hard failure.
```
swennemers, line 19: I am not sure about this change. Did we consider the downstream affect on ORD consumers relying on major versions being in sync? Did we align this with the provisioning CoE. 
If organizations bump major versions, they create an incompatible change. Allowing them to keep the ORD-ID and having only a partial major versions seems hacky. We should have a very clear picture of the top-consumers (e.g. BDC & provisioning CoE) whether relaxing the rule is acceptable.

swennemers (changes requested):

CHANGELOG.md
Fannon: Yes, agreed. But if in practice not every application can follow the versioning / lifecycle opinion we stated (also for historical reasons), I'm not sure if this assumption would break even worse. 

It might be easier to not try to extract out semantics of the ORD ID anyway, but we need to check if that's now done. If for e.g. Data Products we consistently do it correctly via the platform, the platform can also still rely on this.

swennemers (approved):
I am not happy with it, but accept is as pragmatic solution

PR #140, opened 2026-05-28
Make partOfProduct a clearer recommendation

Fannon:
### Changed

- Strengthened `partOfProducts` guidance: every ORD resource SHOULD be assigned to at least one product, either directly or inherited from its package. Setting `partOfProducts` on the `Package` is the preferred approach as it propagates automatically to all contained resources.


spec/v1/Document.schema.yaml
```diff
@@ -692,6 +696,8 @@ definitions:
       - description
       - version
       - vendor
+    x-recommended:
+     - partOfProducts
```
Fannon, line 700:
```suggestion
```

Not sure we should use x-recommended because it will always give a warning, even if the assignment is consistently done on resource level. We could also drop it and implement the check as a custom validation rule which is smarter.

Fannon: Ok, recondsidering: Since package can be in different ORD document, we wouldn't know for sure what's inherited. And it's actually sensible to always do the product assignment on package level, even if the resources are explicitly assigned. So I think going with x-recommended would be easiest and most declarative way to at least warn / hint at potential problems.

spec/v1/Document.schema.yaml
```diff
@@ -533,11 +533,15 @@ definitions:
       partOfProducts: &partOfProducts
         type: array
         description: |-
-          List of products the resources of the Package are a part of.
+          List of products this Package and its resources are a part of.
```
```suggestion
          List of products this package and its resources are a part of.
```

PR #141, opened 2026-05-29
Add ord:ai-hint predefined label key

Fannon:
### Added

- Added `ord:ai-hint` as a pre-defined label key on all ORD resources for guidance targeted at AI consumers (LLMs, orchestrators), kept intentionally separate from human-readable description fields. MUST contain exactly one value; SHOULD be CommonMark Markdown.


Fannon:
Closing in favor of dedicated `aiHint` property, see PR #143

PR #142, opened 2026-06-02
Better Integration Dependency / subset documentation

Fannon:
Since subset is now more important also in the agent -> MCP tools dependency context, this improves the documentation and also adds examples for agents

### DIFF
--- a/spec/v1/Document.schema.yaml
+++ b/spec/v1/Document.schema.yaml
@@ -3721,7 +3721,7 @@ definitions:
           If `subset` is not provided, the dependency implies that all operations of the referenced resource may be used.
           If `subset` is provided, only the listed operations are required — consumers MUST NOT assume that other operations are available or permitted.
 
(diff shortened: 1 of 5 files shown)


spec/v1/Document.schema.yaml
```diff
@@ -3818,10 +3823,15 @@ definitions:
...
+      Defines that the API Resource Integration Aspect only requires a subset of the referenced contract.
 
       For APIs, this is a list of the operations or tools that need to be available in order to make the integration work.
-      This information helps to narrow down what is really necessary and can help optimize the integration.
+      Without a `subset`, the dependency implies access to the full resource.
+      With a `subset`, only the listed operations are required, allowing consumers to understand the minimal surface area needed.
+
+      This is particularly important for agents consuming MCP servers:
+      the listed tools can be selectively loaded into the LLM context to reduce prompt size and cost,
+      and permission grants or routing decisions can be scoped to exactly what is required.
```
```suggestion
      With a `subset`, only the listed operations are required, allowing consumers to understand /load only the minimal surface area needed.
```

PR #143, opened 2026-06-02
Add AI Hint property

Fannon:
### Added

- Added `aiHint` as a dedicated, optional property on API Resources, Event Resources, Entity Types, Data Products, Agents, and Capabilities. Provides guidance targeted at AI consumers (LLMs, orchestrators), kept intentionally separate from human-readable `description` fields so both can evolve independently. SHOULD be CommonMark Markdown. See [AI Agents and Protocols](https://open-resource-discovery.org/spec-v1/concepts/ai-agents-and-protocols#ai-hints-on-ord-resources).

Obsoletes #141

### DIFF
--- a/spec/v1/Document.schema.yaml
+++ b/spec/v1/Document.schema.yaml
@@ -1706,8 +1706,6 @@ definitions:
 
       aiHint:
         <<: *aiHint
-        examples:
-          - "Subscribe to this event to react when an invoice is cancelled. Filter by `companyCode` to narrow to relevant tenants."
 
       ## Assignments to groupings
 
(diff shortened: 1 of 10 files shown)


spec/v1/Document.schema.yaml
```diff
@@ -2182,6 +2204,11 @@ definitions:
         examples:
           - The data product Customer Order offers access to all online and offline orders submitted by customers. It provides a customer view on the orders. For fulfillment-specific aspects please refer to the data product Fulfillment Order.
 
+      aiHint:
+        <<: *aiHint
+        examples:
+          - "Use this data product to access **customer order** data. Filter by `customerId` and `orderDate` for targeted queries. Output port `customerOrderItems` provides line-item detail."
```
Fannon, line 2214: @swennemers can you check if this example makes sense?  Maybe we drop examples where we are not sure? Once we know better what works, we can add better examples.

swennemers: I like the example for a DP (I don't see it in the PR anymore)

swennemers (approved):

spec/v1/Document.schema.yaml
Fannon: Added it back

PR #146, opened 2026-06-04
Fix URL resolution: rename root-relative to base-URL-relative, clarify semantics

### LINKED ISSUE #145: Root-relative URL resolution claim contradicts RFC 3986
## Summary

The [Relative URL Resolution](https://open-resource-discovery.org/spec-v1/#relative-url-resolution) section introduced in ORD 1.16 contains a factually incorrect claim about how root-relative URLs are resolved under RFC 3986.

## The Problem

The spec states (index.md):

> Root-relative URLs (leading slash) are resolved against the applicable base URL *including its path component*.
> For example, if `baseUrl` is `https://provider.com/api/v1`, then `/metadata/schema.json` resolves to `https://provider.com/api/v1/metadata/schema.json`.

This is **incorrect** per [RFC 3986 §5.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.2). A leading-slash reference takes only the scheme and authority from the base URI and discards its path entirely:

```
base: https://provider.com/api/v1
ref:  /metadata/schema.json
RFC 3986 result: https://provider.com/metadata/schema.json
spec claim:      https://provider.com/api/v1/metadata/schema.json  ← WRONG
```

All standard URL libraries implement the RFC 3986 behaviour:
- Python: `urllib.parse.urljoin("https://provider.com/api/v1", "/metadata/schema.json")` → `https://provider.com/metadata/schema.json`
- JavaScript: `new URL("/metadata/schema.json", "https://provider.com/api/v1").href` → `https://provider.com/metadata/schema.json`
- Browser `<a href="/metadata/schema.json">` from `https://provider.com/api/v1` → same

The spec simultaneously cites RFC 3986 and contradicts it. Every implementer that uses a standard URL library will produce results contrary to the spec's example and description.

The resolution table in the same section repeats the same error:
```
| `/metadata/schema.json` | Metadata file | https://provider.com/api/v1/metadata/schema.json |
```

## Fix Options

**Option A (preferred) — align the spec with RFC 3986:**
Change the description to match standard behaviour. Root-relative URLs strip the base path and resolve against scheme+authority only. Update the example table accordingly.

**Option B — define ORD-specific resolution:**
If the intended behaviour is "append ref path to base path", explicitly define this as an ORD-specific rule, remove the RFC 3986 citation for this case, and document that ORD aggregators MUST implement this non-standard resolution. This is not recommended as it diverges from every standard library.

## Affected Files

- `docs/spec-v1/index.md` — URL Reference Types table, "Root-relative URLs" description paragraph, resolution examples table

## Impact

Any ORD aggregator or ORD provider that implements relative URL resolution using a standard URL library will resolve root-relative URLs differently from what the spec claims. This is a correctness bug that will cause interoperability issues.

Fannon:
ORD's leading-slash URL resolution uses baseUrl-path-prefix semantics (baseUrl + "/" + path), not RFC 3986 root-relative (which strips the base path). Rename the type to "Base-URL-relative" to avoid the RFC 3986 naming conflict, add an explicit concatenation rule, and document the no-baseUrl fallback. Document-relative (./  ../) resolution continues to follow RFC 3986 §5 unchanged.

Closes #145

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ## [unreleased]
 
+### Fixed
+
+- Clarified that base-URL-relative URLs (leading `/`) in ORD use **ORD-specific concatenation semantics** (`baseUrl + "/" + path`), not RFC 3986 root-relative resolution (which strips the base path). Renamed the URL reference type from "Root-relative" to "Base-URL-relative" to avoid the RFC 3986 naming conflict. Added explicit concatenation rule and no-`baseUrl` fallback (scheme+authority only, which coincides with RFC 3986 behavior). Document-relative (`./`, `../`) URLs continue to follow RFC 3986 §5 unchanged.
+  See [issue #145](https://github.com/open-resource-discovery/specification/issues/145).
+
 ### Added
 
 - Added `definitions` property to `EntityType` resources to allow linking machine-readable entity type definitions
--- a/docs/spec-v1/index.md
+++ b/docs/spec-v1/index.md
@@ -315,8 +315,6 @@ Please [create an issue](https://github.com/open-resource-discovery/specificatio
 ORD documents may contain both absolute and relative URLs.
 ORD aggregators MUST resolve all relative URLs to absolute URLs before exposing them to ORD consumers.
 
-Relative URL resolution follows [RFC 3986 (Uniform Resource Identifier: Generic Syntax)](https://datatracker.ietf.org/doc/html/rfc3986#section-5), with the base URI determined as described below.
-
 ##### URL Reference Types
 
 ORD recognizes three types of URL references:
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ## [unreleased]
 
+### Fixed
+
+- Clarified that base-URL-relative URLs (leading `/`) in ORD use **ORD-specific concatenation semantics** (`baseUrl + "/" + path`), not RFC 3986 root-relative resolution (which strips the base path). Renamed the URL reference type from "Root-relative" to "Base-URL-relative" to avoid the RFC 3986 naming conflict. Added explicit concatenation rule and no-`baseUrl` fallback (scheme+authority only, which coincides with RFC 3986 behavior). Document-relative (`./`, `../`) URLs continue to follow RFC 3986 §5 unchanged.
+  See [issue #145](https://github.com/open-resource-discovery/specification/issues/145).
```
Fannon, line 16: Since 1.16 is not released yet, this fixed entry can be removed again once merged with the rest of 0.16.0

PR #148, opened 2026-06-05
Add ORD Overlays support for patching ORD information

Fannon:
### Added

- Bumped ORD Overlay spec to `0.2` and reintroduced the `ordId` selector (`OverlaySelectorByOrdId`).
  - A patch can now target an ORD resource itself (e.g. an API Resource or Event Resource) by its [ORD ID](https://open-resource-discovery.org/spec-v1/#ord-id) and patch its ORD-level metadata (title, description, tags, visibility, ...) without going through a referenced definition file.
  - When all patches in an overlay use `selector.ordId`, the document-level `target` MAY be omitted — each patch carries its own resource identifier and a single overlay file MAY patch multiple ORD resources at once.
  - ORD Aggregators MUST apply such overlays when building their ORD Discovery API and related indexes so that ORD-level overlay changes are reflected in discovery, filtering, and searching.
  - The `ordOverlay` enum now accepts both `"0.1"` (kept for backward compatibility with overlays produced under ORD 1.16.0) and `"0.2"` (new overlays SHOULD use `"0.2"`).


spec/v1/Document.schema.yaml
```diff
@@ -2839,7 +2839,7 @@ definitions:
             x-feature-status: alpha
             x-introduced-in-version: "1.15.0"
             description: |-
-              ORD Overlay file for patching referenced resource definition files.
+              ORD Overlay file for patching ORD resources and/or referenced resource definition files.
```
swennemers, line 2842: I don't think we should allow patching ORD resources at this place, next to the resource definitions
```suggestion
              ORD Overlay file for patching referenced resource definition files.
```

spec/v1/Document.schema.yaml
```diff
@@ -2991,7 +2991,7 @@ definitions:
             x-feature-status: alpha
             x-introduced-in-version: "1.15.0"
             description: |-
-              ORD Overlay file for patching metadata / resource definition files.
+              ORD Overlay file for patching ORD resources and/or referenced resource definition files.
```
swennemers, line 2994: see above

spec/v1/OrdOverlay.schema.yaml
```diff
@@ -34,11 +34,18 @@ properties:
...
   ordOverlay:
     type: string
-    description: Version of the ORD Overlay specification.
+    description: |-
+      Version of the ORD Overlay specification.
+
+      Version `0.2` adds the `ordId` selector for patching ORD resource metadata directly,
+      and allows the document-level `target` to be omitted when all patches use `selector.ordId`.
+      Version `0.1` is still accepted for backward compatibility with overlays produced under
+      ORD 1.16.0; new overlays SHOULD use `"0.2"`.
```
swennemers, line 43: sure that we want the changelog in the description? do we have an alternative place for this.

PR #158, opened 2026-07-03
Add Arazzo workflow support (arazzo-v1, workflow capability, ord:orchestrates)

Fannon:
## Summary

Adds support for [Arazzo](https://www.openapis.org/arazzo-specification) multi-step API workflows in ORD.

- **`arazzo-v1` on API Resource `resourceDefinitions`** — for workflows scoped to a single API.
- **`workflow` Capability type** with **`arazzo-v1` capability definition** — for workflows that span multiple APIs. Such capabilities SHOULD list every covered API in `relatedApiResources`.
- **`ord:orchestrates` relationType** — declares that the source resource (e.g. a `workflow` Capability) orchestrates the target API Resource.

Scheduled for the next patch release (1.16.4).

### DIFF
--- a/spec/v1/Document.schema.yaml
+++ b/spec/v1/Document.schema.yaml
@@ -2849,6 +2849,18 @@ definitions:
               ORD Overlay file for patching referenced resource definition files.
               For details, see [ORD Overlay](../../spec-v1/interfaces/OrdOverlay).
               The `mediaType` MUST be set to `application/json`.
+          - const: arazzo-v1
+            x-introduced-in-version: "1.16.4"
+            description: |-
+              [Arazzo Specification v1.0.0](https://spec.openapis.org/arazzo/v1.0.0.html) workflow document
+              describing multi-step API interactions.
+              The `mediaType` MUST be set to `application/json` or `application/yaml`.
+
+              Attaching an Arazzo definition to a single API Resource is ONLY appropriate when the
+              workflow is scoped to that specific API. For workflows that span multiple APIs, publish
+              a [Capability](#capability) with `type: "workflow"` and a definition of `type: "arazzo-v1"` instead,
+              and list every covered API via [`relatedApiResources`](#capability_relatedapiresources) with
+              `relationType: "ord:orchestrates"`.
           - const: custom
             description: |-
               If chosen, `customType` MUST be provided.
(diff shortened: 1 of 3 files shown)


spec/v1/Document.schema.yaml
```diff
@@ -2849,6 +2849,18 @@ definitions:
...
+            description: |-
+              [Arazzo Specification v1.0.0](https://spec.openapis.org/arazzo/v1.0.0.html) workflow document
+              describing multi-step API interactions.
+              The `mediaType` MUST be set to `application/json` or `application/yaml`.
+
+              Attaching an Arazzo definition to a single API Resource is ONLY appropriate when the
+              workflow is scoped to that specific API. For workflows that span multiple APIs, publish
+              a [Capability](#capability) with `type: "workflow"` and a definition of `type: "arazzo-v1"` instead,
+              and list every covered API via [`relatedApiResources`](#capability_relatedapiresources) with
+              `relationType: "ord:orchestrates"`.
```
pavelkornev, line 2863: Can always be a capability then, or?

spec/v1/Document.schema.yaml
```diff
@@ -5513,6 +5543,13 @@ definitions:
             description: |-
               The source resource patches one or more definition files of the target resource.
               Used on [ORD Overlay Resources](#overlay) to declare which API or Event resource they patch.
+          - const: "ord:orchestrates"
```
pavelkornev, line 5546: >orchestrates

I think we should use "noun" as type names -> "orchestrator"


Can't it be also called `ord:workflow` here? Do we really need to introduce a new term?

src/generated/spec/v1/types/Document.ts
```diff
@@ -913,7 +913,7 @@ export interface RelatedAPIResource {
    * Defines the semantic meaning of the relationship.
    * If not provided, the relationship has no specific semantics ("related somehow").
    */
-  relationType?: (string | "ord:patches") & string;
+  relationType?: (string | "ord:patches" | "ord:orchestrates") & string;
```
pavelkornev, line 916: We also need to add "ord:alias" here (it's already in-use)

spec/v1/Document.schema.yaml
swennemers: actually, we use verbs here  x "patches" y

spec/v1/Document.schema.yaml
swennemers: agree, the decision on API vs. Capability is unclear. How often to we have the use case of one API only arazzo definition? Do we need the complexity of adding it to API and as capability?

src/generated/spec/v1/types/Document.ts
Fannon: Ok, separate PR or include it here?

spec/v1/Document.schema.yaml
Fannon: Agree with Sebastian, I would prefer consistent verb use

spec/v1/Document.schema.yaml
Fannon: Could be yes. But from what I've seen it doesn't seem uncommon to have a complementary Arazzo file for exactly one API, describing the flows within that one API. In this case it would be more convenient to have it directly attached to the API. 

On the other hand, now we have two places where Arazzo could show up, which adds complexity as well.

Leaning toward making it a capability only

PR #170, opened 2026-08-31
feat(spec): add vendor and product assignments to ORD resources

Fannon:
## Summary

Adds optional `vendor` and `partOfProducts` properties directly to every ORD resource type.
Package-level inheritance remains supported.
Direct resource-level `partOfProducts` assignments are complementary and are merged with inherited Package assignments.

## Motivation

This prepares the model for making Package assignment optional and potentially deprecating Packages in the future.
The related draft proposal to make `partOfPackage` optional is https://github.com/open-resource-discovery/specification/pull/162.
Moving the remaining shared metadata to resources lets consumers interpret and validate a resource without first resolving its Package, which may be published in another ORD document.
This is also needed by policies such as `sap:ai:v1`, which must determine the responsible vendor while validating an individual resource.

## Package-independent resources

A resource without a `partOfPackage` assignment cannot inherit Package metadata and must directly provide every property required from its Package context.
In the current model, `vendor` is the relevant mandatory Package property that was not previously available on every resource, while common identity and descriptive fields are already present directly on resources where applicable.
The specification now states that `vendor` MUST be provided directly for Package-capable resource types when no Package is assigned.
When draft PR 162 removes the unconditional `partOfPackage` requirement, its validation must enforce the resulting condition that either `partOfPackage` or direct `vendor` is present.

Product assignment remains recommended rather than mandatory.
When a resource has no Package, any product association must be expressed through its direct `partOfProducts` property.

## Open discussion

As an alternative, `vendor` and `partOfProducts` could be defined as ORD document-level defaults inherited by resources.
This could reduce repetition, but it would require clear override and merge semantics and must support documents containing resources from different vendors or products.
This PR does not introduce document-level defaults.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ### Changed
 
+- Added `vendor` directly to all ORD resources as an optional property.
+  When a contained resource omits it, ORD Aggregators SHOULD inherit the `vendor` assigned to its Package.
 - Clarified `OverlaySelectorByOperation` for OData targets: when the operation name alone is ambiguous (e.g. bound operations overloaded on multiple entity types), the selector MUST use the fully qualified signature (including parameters, e.g. `OData.Demo.Approval(Edm.Int32,Edm.String)`) or fall back to `jsonPath` to target the specific overload. When the name is already unique, the plain name remains sufficient. This makes explicit the pre-existing requirement that operation selectors resolve unambiguously.
 - Updated the ORD Overlay Tooling section to link the published reference implementation (`overlay-tools`), a set of React components for viewing Overlay documents (`overlay-editor`), and a reusable Go library (`overlay-golang`).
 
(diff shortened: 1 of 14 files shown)


Fannon:
Opening this for discussion - I'm not sure if it really makes sense to override the vendor on per-resource-level. 

Ideally this is handled by the aggregator - the "vendor" of a resource could be a calculated, read-only property on the Aggregator Discovery API, but an ORD Provider wouldn't need to explicitly describe it in the ORD Documents.

Right now ORD spec doesn't define the Aggregator Discovery API, this makes it a bit difficult to address. I think it would be better not to add the property as suggested here, but to let the aggregator discovery API forward it, so consumers can filter more easily. 

For validation it's tricky, because one ORD document may not include the package that inherits the information down to the resource level, so we can't reliably use it, e.g. for conditional validation.

Fannon:
Superseded by #175, which makes resource-level vendor explicitly RECOMMENDED, includes Overlay resources, and clarifies the sap:ai:v1 isolated-validation motivation.

PR #171, opened 2026-08-31
docs(overlay): clarify selector and OData semantics

Fannon:
## What

Makes previously implicit or inconsistent ORD Overlay v1 alpha behavior explicit.
There are no schema-shape changes.
The normative changes are descriptions in `spec/v1/OrdOverlay.schema.yaml` and `OrdOverlay.intro.md`, with regenerated artifacts and a changelog entry.
The overlay wire value remains `0.1` because the model is still alpha and only major/minor model versions are accepted.

## Clarified contract

- Unmatched concept-level `merge` and `update` selectors MUST error and MUST NOT create missing structure.
- Unmatched `remove` is a successful idempotent no-op, and tooling SHOULD warn.
- A `jsonPath` selector may match zero or more elements, applies a patch exactly once to every location identified before mutation, and produces a warning-only no-op when it matches nothing, regardless of action.
- JSONPath evaluation follows RFC 9535 with a mandatory portable profile: root, dot and quoted-bracket member selectors, non-negative indices, object and array wildcards, homogeneous selector lists, and non-negative forward slices.
- Portable overlays MUST stay within that profile.
  Toolkits MAY support non-portable extensions, while unsupported expressions MUST error rather than appear as zero matches.
- Missing fields inside a removal mask are ignored.
- Omitted-data removal of the document root MUST error, while root removal masks remain valid.
- OData annotation identity is term plus optional qualifier.
- EDMX overlays preserve structural elements and have explicit `merge`, `update`, and `remove` annotation-set semantics.
- OData v4 EDMX output uses reconciled external `<Annotations Target="...">` blocks.
- Applying overlays to OData v2 EDMX is unsupported and MUST error.
  An OData v4 EDMX document may describe the model of an OData v2 API and serve as the overlay target.
- External target identifiers cover namespaces, types, members, containers, operations, parameters, and return types.
- OData operation selectors use exact signatures when overloaded, including declaration order, binding parameters, and explicit `Name()` for zero parameters.
- Signature selectors never fall back to `FunctionImport`, and parameter and return-type selectors use the same operation syntax.
- CSDL JSON concept-level `merge`, `update`, and data-bearing `remove` operate on annotations while preserving structural content.
- CSDL JSON concept-level `update` replaces the selected target's complete annotation set, while `root` and `jsonPath` retain generic whole-element semantics.
- Structural CSDL data supplied through a concept-level selector is an error.
- CSDL JSON enum-member annotations use sibling keys with explicit merge, update, and remove behavior.
- Output equivalence is semantic rather than byte-for-byte.

## Toolkit processing policy

The specification defines patch outcomes, not toolkit control flow or CLI presentation.
Collecting diagnostics across patches, output streams, process exit codes, and optional warning escalation remain implementation concerns.
The toolkit PRs aggregate errors, continue processing, and return a failing outcome after all patches have been attempted.
An `--error-on-warnings` option is intentionally deferred until the diagnostic interfaces stabilize.

## Toolkit alignment and merge order

Merge this specification PR before the corresponding toolkit behavior and documentation PRs.

- TypeScript: [#68](https://github.tools.sap/ORD/overlay-tools/pull/68) first; [#70](https://github.tools.sap/ORD/overlay-tools/pull/70) and [#72](https://github.tools.sap/ORD/overlay-tools/pull/72) after #68; [#74](https://github.tools.sap/ORD/overlay-tools/pull/74) after #72; [#73](https://github.tools.sap/ORD/overlay-tools/pull/73) is independent; [#76](https://github.tools.sap/ORD/overlay-tools/pull/76) follows #73; the portable-profile tests in [#77](https://github.tools.sap/ORD/overlay-tools/pull/77) follow #76; [#71](https://github.tools.sap/ORD/overlay-tools/pull/71) owns aggregate documentation and the changelog and merges last.
- Go: [#9](https://github.com/open-resource-discovery/overlay-golang/pull/9), [#10](https://github.com/open-resource-discovery/overlay-golang/pull/10), [#16](https://github.com/open-resource-discovery/overlay-golang/pull/16), [#20](https://github.com/open-resource-discovery/overlay-golang/pull/20), and [#21](https://github.com/open-resource-discovery/overlay-golang/pull/21) are independent; [#14](https://github.com/open-resource-discovery/overlay-golang/pull/14) and #9 precede [#17](https://github.com/open-resource-discovery/overlay-golang/pull/17); the portable-profile tests and multi-match removal fix in [#18](https://github.com/open-resource-discovery/overlay-golang/pull/18) follow #17; [#15](https://github.com/open-resource-discovery/overlay-golang/pull/15) implements the CSDL JSON annotation-set update direction but still needs to reject structural patch data; [#8](https://github.com/open-resource-discovery/overlay-golang/pull/8) owns aggregate documentation and the changelog and merges last.


spec/v1/OrdOverlay.intro.md
```diff
@@ -143,16 +147,47 @@ The full semantics of each action (`update`, `merge`, `remove`) are defined on t
...
+  A future version may add a dedicated create action.
+  See [`selector`](#overlay-selector).
+- **OData targets**: overlays carry annotations in CSDL JSON `@TermName` form.
+  EDMX targets are annotation-only: existing structure is annotated, new structural elements cannot be created, and `jsonPath` is unavailable for EDMX.
+  OData v4 EDMX annotations use reconciled external `<Annotations Target="...">` blocks.
+  OData v2 EDMX annotations are written inline using syntax supported by the target document, and tooling reports an error when no meaning-preserving representation exists.
+  For CSDL JSON, enum-member annotations are sibling keys on the enum type.
+  Updating an enum member preserves its scalar value and replaces its sibling annotations.
+  Removing an enum member without `data` removes both the member and its sibling annotations.
+  See [`data`](#overlay-patch-value).
```
Fannon, line 183: @mlakov - this would mean that we have to also support OData v2 with golang explicitly, I guess

spec/v1/OrdOverlay.intro.md
```diff
@@ -143,16 +147,48 @@ The full semantics of each action (`update`, `merge`, `remove`) are defined on t
...
+    The document root cannot be removed: `root`, or `jsonPath: "$"`, with omitted `data` MUST error.
+    Use `update` to replace the complete document explicitly.
+    For EDMX, the selected structural element is preserved and its complete annotation set is removed.
   - Provide `data` with `null`-valued properties to remove only those specific fields.
-  - `data` MUST NOT be `null`, an empty object `{}`, or an empty array `[]` — these are invalid and will be rejected by conformant tooling.
-- **`merge` behavior**: arrays are appended, not replaced. To fully replace an array, use two ordered patches — first `remove` the array field with `data: { "arrayField": null }`, then `merge` the new value.
+    Missing fields in a matched element are ignored.
+    A removal mask MAY target the document root because it preserves the document itself.
+  - `data` MUST NOT be `null`, an empty object `{}`, or an empty array `[]`; conformant tooling rejects these values.
+- **`merge` behavior**: arrays are appended, not replaced.
```
mlakov: Isn't it correct to use `update` instead ?

spec/v1/OrdOverlay.intro.md
```diff
@@ -143,16 +147,48 @@ The full semantics of each action (`update`, `merge`, `remove`) are defined on t
...
+  To replace an array, first `remove` it with `data: { "arrayField": null }`, then `merge` the new value.
+- **CSDL JSON concept selectors**: `merge`, `update`, and data-bearing `remove` operate on annotations while preserving structural CSDL content.
+  `update` replaces the selected target's complete annotation set.
+  This exception does not apply to `root` or `jsonPath`, which retain generic structural semantics.
+  Structural CSDL data supplied to an OData concept-level selector is an error.
+- **Selector matches**: a `merge` or `update` concept-level selector that matches nothing is an error.
```
mlakov, line 166: The current implementation in the Golang library is such that it will error out if you attempt a `remove` with a concept-level selector for a missing element. Shall I change that, @Fannon ?

spec/v1/OrdOverlay.schema.yaml
```diff
@@ -928,9 +994,47 @@ definitions:
...
 
+      EDMX targets are annotation-only: overlays add, replace, or remove annotations on structure that
+      already exists in the source schema. They cannot create new structural elements (EntityType,
+      EntitySet, Property, Function/Action, EnumType, ComplexType), and `jsonPath` is not available for
+      EDMX. New structure belongs in the source CSDL/CDS, not in an overlay.
+      For EDMX, `merge` replaces annotations with the same term-plus-qualifier identity and preserves
+      all others; `update` replaces the selected target's complete annotation set; and `remove` with
+      omitted `data` removes that complete annotation set while preserving the structural target.
+
+      For CSDL JSON, enum-member annotations are carried as sibling keys on the enum type
```
mlakov, line 1021: The current Golang implementation does not support specifying annotations specified like that. Support for annotating enum type members is based on the same syntax as with entity type or complex type:

```
...
    {
        "Read": {
            "@Core.Description": "<description>"
        }
    }
...
```

Should I be changing this, @Fannon  ?

spec/v1/OrdOverlay.schema.yaml
```diff
@@ -584,13 +639,19 @@ definitions:
...
-            when no Schema-level Action/Function matches the name.
-            When the name alone is ambiguous (e.g. bound operations overloaded on multiple entity types),
-            the selector MUST use the fully qualified signature (incl. parameters, e.g.
-            `OData.Demo.Approval(Edm.Int32,Edm.String)`), or fall back to [`jsonPath`](#overlay-selector-by-jsonpath),
-            to target the specific overload. When the name is already unique, the plain name is sufficient.
+            when a selector without a signature matches no Schema-level Action or Function.
+            A signature selector MUST NOT match a FunctionImport.
+
+            A namespace-qualified name without a signature is sufficient only when it identifies
+            exactly one Action or Function.
```
mlakov: `Action, Function or Function Import` ?

spec/v1/OrdOverlay.intro.md
mlakov, line 159: Isn't it correct to use update instead ?

spec/v1/OrdOverlay.intro.md
Fannon: For remove, let's not create an exception, we should not error on missing element, be it JSON path or semantic selector

spec/v1/OrdOverlay.schema.yaml
Fannon:
```suggestion
            exactly one Action, Function or Function Import.
```

spec/v1/OrdOverlay.schema.yaml
Fannon: Hm, I would need to understand this better. 

I'm not sure if we need this at the moment, if we're unsure, we could also keep this out of scope for now? Maybe not mention it for now and the toolkit just errors if something unsupported is asked for. 


```suggestion
```

spec/v1/OrdOverlay.intro.md
Fannon: We decided to drop OData 2 EDMX support entirely (use OData v4 EDMX to describe v2 instead). This resolves that problem.

spec/v1/OrdOverlay.intro.md
Fannon: Yes, when the selector targets the array itself, `update` is correct. I confirmed that both implementations replace the selected node for `update` and append arrays for `merge`; Go also has an integration test that replaces the selected `$.servers` array. I clarified the text to distinguish this from replacing an array-valued field while preserving the rest of its containing object.

spec/v1/OrdOverlay.schema.yaml
mlakov: From my point of view this is the correct approach for handling enum type member annotations. 

What made me consider an alternative is the fact that for all other cases (e.g. entity type properties) the syntax is different. This may lead to end user errors eventually.

spec/v1/OrdOverlay.intro.md
mlakov: Understood & agreed. I’ll create a follow up change in the Golang implementation

spec/v1/OrdOverlay.schema.yaml
Fannon: Ok, I guess we'll leave it at that, but if you see any further issues, feeel free to bring it up or propose any other clarifications. Thanks a lot, @mlakov

PR #175, opened 2026-09-08
feat(spec): add recommended vendor to ORD resources

Fannon:
## Summary

Adds `vendor` directly to every ORD resource as a RECOMMENDED property.
This includes API Resources, Event Resources, Entity Types, Data Products, Agents, Overlay Resources, Capabilities, and Integration Dependencies.

When a contained resource omits `vendor`, an ORD Aggregator SHOULD inherit the vendor assignment from its Package.
The generated TypeScript and UMS artifacts are updated accordingly.

## Motivation

Resource-level consumers and validators cannot always resolve `partOfPackage.vendor`, because the referenced Package may be provided in another ORD document.
In particular, the `sap:ai:v1` policy needs an explicit resource vendor so isolated validation can determine whether `vendor` is `sap:vendor:SAP:` before requiring the standardized Jarvis ID.
The public ORD requirement remains RECOMMENDED for compatibility, while `sap:ai:v1` can make explicit `vendor` mandatory.

Supersedes #170.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ For a roadmap including expected timeline, please refer to [ROADMAP.md](./ROADMA
 
 ### Changed
 
+- Added `vendor` directly to all ORD resources as a recommended property.
+  When a contained resource omits it, ORD Aggregators SHOULD inherit the `vendor` assigned to its Package.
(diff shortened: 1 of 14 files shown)


Fannon:
Superseded by the refined existing PR #170. The updated #170 keeps vendor optional in the general ORD specification and documents the remaining sap:ai:v1 policy choice.

PR #176, opened 2026-09-10
docs: clarify Customer Namespace and reserve `c` vendor alias

Fannon:
## What

Clarifies the Customer Namespace section and reserves a short `c` alias for the `customer` vendor namespace.

## Why

The existing guidance does not clearly connect the customer namespace with the Package vendor assignment used to identify customer-owned content.

## Changes

- State that customer-created resources MUST be assigned to a Package whose `vendor` is `customer:vendor:Customer:`.
- Reserve `c` as a shorter alias of `customer` with identical semantics.

## Companion change

The corresponding registry entries for the `c` alias and its mapping to `customer:vendor:Customer:` were added in ORD/global-registry#1171.


docs/spec-v1/index.md
```diff
@@ -870,7 +871,14 @@ To keep this situation simple, there is a reserved [vendor namespace](#vendor-na
...
 
-The limitation of using `customer.*` namespaces is that they are unique only within a tenant and once the resources are published and shared outside the local scope, the `customer` namespace will be insufficient.
+Resources that are created by the customer / user MUST be assigned to a [Package](#package) whose `vendor` is set to the reserved customer vendor `customer:vendor:Customer:`.
+This is how consumers recognize customer-owned content, independent of which namespace prefix the individual resource IDs use.
+
+For cases where the [36-character namespace length limit](#namespace-constraints) is tight, a shorter alias `c` is also reserved and is equivalent to `customer`.
+Both `customer.*` and `c.*` map to the same customer vendor `customer:vendor:Customer:`.
+A [`Vendor`](#vendor) MAY list the vendor namespaces that map to it under the reserved `ord:namespace` label, as done for the customer vendor.
+
+The limitation of using `customer.*` (or `c.*`) namespaces is that they are unique only within a tenant and once the resources are published and shared outside the local scope, the `customer` namespace will be insufficient.
```
swennemers, line 881:
```suggestion
The limitation of using `customer.*` (or `c.*`) namespaces is that they are unique only within a tenant and once the resources are published and shared outside the local scope, in a `system version` or `system type` perspective, the `customer` namespace will be insufficient.
```

swennemers: @Fannon I think, we should make it more explicit that system version + system type is not allowed.

swennemers (commented):
in general I am fine, see comment

docs/spec-v1/index.md
Fannon: @swennemers I wonder about this, too. Either you pick system-type or system-version, depending on whether you need versioning or not. ?

This would be a more general clarification on the perspectives and go beyond the scope of this PR.

swennemers: why do you see it out of scope, I just proposed, we should sharpen the description here or do you see the customer namespace outside system-instance perspective?

swennemers: @Fannon , you misunderstood my comment. I meant that we should for make it explicit that the customer namespace (and its alias) are not allowed in either system-version or system type perspective, not just that is is insufficient.

PR #179, opened 2026-09-15
fix: clarify static perspective publication and resolution

Fannon:
## Summary

- treat `system-type` and `system-version` as alternative static publication models for a given system type
- use the latest `system-version` as the current static view when versioned metadata exists, and otherwise use `system-type`
- keep exact-version requests exact and keep `system-instance` as a separate, coexisting dynamic perspective
- align the specification, perspective concept page, and schema descriptions

## Why this is a bug fix

The current documentation recommends choosing `system-type` for unversioned metadata and `system-version` for versioned metadata, but it does not prohibit publishing both.
Its resolution guidance also gives an explicit `system-type` perspective precedence over the latest `system-version`.
That makes the meaning of the current static view ambiguous and can return stale or contradictory metadata.

## Proposed semantics

- A provider uses either `system-type` or `system-version` for the static metadata of one system type at a given time.
- A request for a specific system version returns only that exact `system-version`.
- A request for the current static view returns the latest `system-version` when available, otherwise `system-type`.
- `system-instance` can coexist with either static model and remains a complete dynamic view.

## Consequences

- Providers that currently publish both static perspectives for one system type need to select one model.
- Aggregators need to determine the latest system version using Semantic Versioning precedence.
- Consumers that currently prefer an explicit `system-type` view over available versioned metadata would instead receive the latest `system-version`.
- A transition between the two models can temporarily expose both perspectives, so migration behavior may need explicit guidance or validation.

## Alternatives to discuss

1. Keep both perspectives valid and retain explicit `system-type` precedence.
2. Keep both perspectives valid but give the latest `system-version` precedence.
3. Make the models mutually exclusive as proposed and define how aggregators handle a temporary overlap during migration.

## Open questions

- Should mutual exclusivity be a `MUST`, or should it initially be a `SHOULD` for compatibility?
- During migration, should an aggregator prefer the latest `system-version`, reject the ambiguous static view, or expose both without resolving a current view?

## Related work

This complements #172.


Fannon:
@swennemers I added the clarification that system-type and system-version should be mutually exclusive to this PR. You brought it up in #176 , but I think the fix / clarification belongs here. 

@desislavaa , @mlakov please check if this makes sense to you, too

Fannon:
One possible simplification is to make the ORD Provider that owns the system onboarding and identity at the ORD aggregator authoritative for the static publication model.
Any delegated ORD Provider publishing metadata on behalf of that system identity would then have to use the same static perspective.

For a versioned system, delegated providers would publish `system-version` metadata for the same concrete versions used by the authoritative provider.
For an unversioned system, they would publish `system-type` metadata.
This removes the need to compose `system-type` and `system-version` contributions for the same system and avoids implicitly turning `system-version` into a delta perspective.
The tradeoff is that a version-independent delegated contribution must be published for every supported concrete system version.

Going one step further, we could consider deprecating and eventually removing `system-type` completely.
Every static ORD publication would then use `system-version`, including continuously delivered systems.
Such systems would need a concrete version identifying an immutable publication or deployment state, even if it is not a commercial product version.

In that model, `latest` would be a movable alias for a concrete system version rather than another perspective.
The concrete `describedSystemVersion.version` would remain an immutable SemVer value.
An optional alias could be represented separately, for example:

```json
{
  "describedSystemVersion": {
    "version": "3.0.0",
    "aliases": ["latest"]
  }
}
```

Suggested resolution rules:

1. A tenant-aware request always resolves to the tenant's exact concrete system version.
2. A request for a specific version always resolves to that concrete version.
3. An unversioned static request resolves to the version explicitly carrying the `latest` alias.
4. If no explicit `latest` alias exists, the aggregator assigns it to the greatest available version according to SemVer precedence.
5. At most one version per system type may carry `latest`.
6. Only the authoritative provider may explicitly assign `latest`.
7. Delegated providers publish against concrete versions and do not control the alias.
8. Conflicting explicit `latest` assignments are validation errors and must not be resolved silently.

This would separate the concepts cleanly:

```text
system-version = concrete immutable static state
latest         = default concrete system-version
```

An explicit alias permits the authoritative provider to designate an older stable or generally available release as `latest`.
Automatic SemVer selection provides deterministic behavior when the provider does not manage the alias.
Removing `system-type` would be a breaking change, so a practical migration could first deprecate it, recommend `system-version` for all new providers, and remove it in the next major ORD version.

docs/spec-v1/concepts/implementing-ord-natively.md
```diff
@@ -92,10 +92,11 @@ The static document describes what a system type or version provides in general
 The dynamic document describes what a concrete system instance provides at runtime, including tenant-specific state — it is fetched per system instance.
 Providers advertise these separately in the ORD configuration so that aggregators can handle them differently.
 
-The `system-version` document should be complete for that version of the application.
-It must not require tenant context and should not contain tenant-specific customizations.
-It should include `describedSystemVersion.version` when the application has a meaningful version.
-If the application is not versioned, consider using `system-type` perspective instead.
+Each resource representation in a `system-version` document should be complete for that version of the application.
```
Fannon: This is a bit confusing, each resource in ORD is anyway described as a unit and therefore always complete. 

But the system-version layer may not be fully complete, for static we can also look on system-type or even system-independent if if's not found on that perspective.

PR #183, opened 2026-09-17
Add extensibility to Agents, Capabilities, and Data Products

Fannon:
## Summary

- Add the optional `extensible` property to Agent, Capability, and Data Product resources for ORD 1.17.
- Generalize `Extensible` wording so it covers custom fields, entities, skills, and other resource-specific extension elements.
- Add validating examples and regenerate the TypeScript and UMS artifacts.

## Scope

Data Product was the other current resource with the same gap, including an existing schema section labeled “Links & Extensibility” that did not expose the property.
Integration Dependency and Overlay are excluded because they describe dependency and patch metadata rather than an extendable contract or functional surface.
The proposed Schema type in https://github.com/open-resource-discovery/specification/pull/163 is also missing `extensible`; it should be added within that PR because Schema is not yet on `main`.

## To Be Clarified
- Also apply it for Data Products? @swennemers input needed


Fannon:
@swennemers Not sure on Data Products, do you have a take on it?

